PR #127, opened 2021-10-06
Refactor download button code

valentinoli:
* Let BaseDownloadButton take href as prop and handle the logic from that point
* Create BaseDataDownloadButton as a higher level component wrapper around BaseDownloadButton to handle downloads of data passed as string
* Remove UrlDataDownloadButton (new BaseDownloadButton is a replacement)


ffsinger:
Nice, it's cleaner like this !

PR #139, opened 2021-10-13
Consent form

ffsinger:
As specified in https://github.com/hestiaAI/data-pipelines/issues/28, this pull request adds the following features:

- Add a title to every experience
- Possibility to design a consent form in the config of a deployment. A form can contain multiple checkboxes or radio button groups.
- If specified, the consent form is shown at the end of the default view, along with checkboxes allowing the user to choose which results to include.
- When the form is filled, possibility to download it as an encrypted zip file.
- A new page `/import` can be used by @pdehaye and @alexbfree to generate cryptographic keys and recreate the results (graphs, tables, and consent log) stored in an encrypted zip.

TODO:

- Implement send button (via netlify forms)

Limitations:

- The created zip is not backward compatible: if an experience is changed, the import may fail.

Important note:

- The import tool will always show the table, even if it was not shown shown to the user. This behavior can be changed if preferred.


andreaskundig:
When I run it with 
```
CONFIG_NAME=workshop npm run dev
```
I get an error after I click on **Process files**

Issue #110, opened 2021-09-20
Change link of the hestia logo to point to the main hestia page


streitlua:
Right now, the hestia logo points to the home page of the experiences tool.
We should probably have this logo to point to the hestia.ai main page, and have a "home" icon to go back to the main experiences page.

valentinoli:
If this is desired behavior then I think having the logo point to hestialabs.org makes more sense since the logo is the one of HestiaLabs.

valentinoli:
I'm not sure if this really is an issue. It is very customary and intuitive from a user's perspective that clicking the logo in the app header takes you to the home page of the app itself.

@streitlua

PR #159, opened 2021-10-20
feat: main logo links to HestiaLabs site (#110)

valentinoli:


streitlua (commented):
I tested the branch on dev and indeed the links work well, I'm just not sure about the home button being on the right (it may be non-intuitive for users, especially when it appears close to the advanced view toggle, which users may perceive as intimidating).
If you don't think that this is an issue, feel free to merge the PR :)

streitlua:
Valentin, I am making changed to the app bar on another branch, so I integrated the changes you've made here on the other branch if that's okay.

Issue #128, opened 2021-10-04
Make query results filterable


andreaskundig:
Add a search field above each column to filter the rows of the table.

Currently we have a search field that searches all columns.

We are using [vuetify's data table](https://dev.vuetifyjs.com/en/components/data-tables/#search). A first glance at the documentation doesn't seem to mention a way to add such a feature. Should we use another component?

andreaskundig:
Meanwhile the visualization components sometimes show several graphs and allow the user to filter the data by clicking on the graphs. It makes less sense now to have a the same filter for visualizations and graphs.

We should ask Alex if what we currently have is enough

PR #160, opened 2021-10-20
feat: query results filter (#128)

valentinoli:
Now the user can... 
* filter the results
* choose which columns to filter

The user still cannot...
* choose how to filter each column


ffsinger (approved):
Nice !

I see that you also added `rel="noreferrer noopener"`, which is good

PR #166, opened 2021-10-21
feat: allow sparql pipelines and custom pipelines to coexist

ffsinger:
Allow sparql and custom pipelines to coexist in a single card and unify twitter and twitter-dashboard.


ffsinger:
Conflicts have been solved. Merging now.

Note: at the moment, only one custom pipeline per card can be defined (e.g. for twitter we have the custom pipeline for the dashboard, and the rest must be sparql). This will be solved later.

PR #178, opened 2021-11-01
fix: graphs not updated when switching file (#173)

### LINKED ISSUE #173: Data not updated when using customPipeline
When using customPipeline, the data is not reloaded if we change the files.

Steps to reproduce: 
   1. Go to https://experiences.hestialabs.org/twitter-dashboard
   2. Load the visualisation with twitter-sample.zip
   3. Load the visualisation with twitter.zip

ffsinger:
The graphs component did not call the `drawViz` function when the values change. It's fixed by using a `watch` property. Keep in mind for future graph components.

Close #173

### DIFF
--- a/components/graphs/TrackerGraph.vue
+++ b/components/graphs/TrackerGraph.vue
@@ -105,6 +105,11 @@ export default {
       results: []
     }
   },
+  watch: {
+    values(oldValues) {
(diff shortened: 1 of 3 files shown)


fquellec:
Thanks for the fix, I'll keep it in mind!

PR #179, opened 2021-11-01
feat: Created new component allowing to explore the given file archive

streitlua:
The related issue is here https://github.com/hestiaAI/digipower-data-experiences/issues/26.

The main addition is the component `UnitFileExplorer.vue`:
- On the left panel, it shows a file directory tree with the contents of the file or list of files that were "uploaded" by the user.
- Clicking on a file on the left panel will result in its visualization on the right panel if the format is supported.
- Currently, the supported file types are json, csv/tsv, pdf, html, images with format png/jpg/jpeg/gif/bmp, and txt. Each file type has a component with dedicated logic.

Notes: 
- To create a tree structure from the files returned by Uppy, I had to add a parallel data processing pipeline, so some processing is duplicated. To avoid that, we would have to completely change the way the processing is done in `process-files.js` and `TheDataExperienceDefault.vue`, while making sure that the old code is compatible, which I found to be very hard and time-consuming.
- I don't know if we want to visualize html files, it could potentially cause security issues, but the whatsapp archive shows most information in an html file so I thought it could be useful.
- To test that everything works correctly, I am mainly testing the experience on twitter-dashboard, but I upload a custom zip file that contains a `data` directory with the expected `.js` files, and with other folders / zips / files. I used private files so maybe it's better for each tester to create their test archive.
- I don't know how to handle the size of the visualizations on the right panel, any suggestions?
- Currently, the generic file viewer is shown for every experience, but I don't know if it's what we want. We could want to control this in the manifest. Also, we could want an experience dedicated to the exploration of the files, to be used when we don't have a dedicated experience for some service.


components/UnitCsvViewer.vue
```diff
@@ -0,0 +1,59 @@
+<template>
+  <div v-if="loading">Loading</div>
+  <v-data-table v-else v-bind="{ ...csvContent }" multi-sort />
```
ffsinger, line 3: Could be better to use the component UnitQueryResults instead of a plain v-data-table. This would provide a filter bar.

Note: in this case we would rename UnitQueryResults into something more fitting (e.g. UnitFilteredTable)

components/UnitCsvViewer.vue
```diff
@@ -0,0 +1,59 @@
...
+      csvContent: {},
+      loading: true
+    }
+  },
+  watch: {
+    async file(newFile) {
+      await this.getContentFromFile(newFile)
+    }
+  },
+  async created() {
```
ffsinger, line 33: We can use use `immediate: true` in the `watch` property above, instead of defining a `created` property that does the same thing.

ffsinger (approved):
From my understanding we want to be able to put files from **any** source, even if no card was created for it. So we would still need to add a **separate** card on the homepage, called "Generic viewer" or something. It would not contain any experience, just the ability to upload a zip and browse the files with this viewer.

components/UnitCsvViewer.vue
streitlua: > From my understanding we want to be able to put files from **any** source, even if no card was created for it. So we would still need to add a **separate** card on the homepage, called "Generic viewer" or something. It would not contain any experience, just the ability to upload a zip and browse the files with this viewer.

Yes, that's what I understand as well. In this case, the problem is that each experience "knows" what files to look for, and currently if this is not specified the code crashes.

streitlua: > Could be better to use the component UnitQueryResults instead of a plain v-data-table. This would provide a filter bar.
> 
> Note: in this case we would rename UnitQueryResults into something more fitting (e.g. UnitFilteredTable)

Yes indeed that seems like a good idea, although I think there is still benefit to having a dedicated `UnitCsvViewer` component, as it avoids complexifying `UnitFileExplorer`

components/UnitCsvViewer.vue
streitlua: That's exactly what I needed, I'll change it!

streitlua: Done, thanks!

Issue #181, opened 2021-11-01
Create experience dedicated to Generic Data Viewer 


streitlua:
Right now the UnitFileExplorer is shown for every experience.

We likely also want to have an experience that is not attached to any service name, which only serves the purpose of understanding the data from some service that we do not support.

We will have to change how the manifests are tested since for this experience we won't be able to give a list of file paths to look for (since it should work for any files).

PR #184, opened 2021-11-02
Feat: created standalone experience for data exploration

streitlua:
Addressing #181 
Also, moved unit-file-explorer to the bottom of the page for the other experiences
Note: the icon comes from here https://materialdesignicons.com/


components/TheDataExperienceDefault.vue
```diff
@@ -143,12 +145,14 @@ export default {
     handleRdfEnd() {
       this.progress = false
     },
-    async onUnitFilesUpdate({ inputFiles, error, allFiles }) {
+    onUnitFilesUpdate({ inputFiles, error, allFiles }) {
       this.initState()
       this.inputFiles = inputFiles
       this.allFiles = allFiles
-      if (this.selectedExample.yarrrml) {
```
andreaskundig, line 150: Leaving this is out might be what breaks the the sparql experiences.

andreaskundig (changes requested):
The sparql experiences are not working. If you choose tinder, process the files and run any graph, it just says "No relevant data found". 

This is probably related to the deletion of a call to parseYarrrml.

components/UnitIntroduction.vue
```diff
@@ -1,7 +1,10 @@
 <template>
   <div>
     <h1>Analyze your private data</h1>
-    <p class="body-1">
+    <p v-if="isGenericViewer" class="body-1">
```
andreaskundig, line 4: We should have a more generic solution for adding a experience specific text, like taking it from a description field in the experience's json

components/TheDataExperienceDefault.vue
streitlua: > The sparql experiences are not working. If you choose tinder, process the files and run any graph, it just says "No relevant data found".
> 
> This is probably related to the deletion of a call to parseYarrrml.

Indeed, this was a mistake I did while merging with the most recent version of master, thanks for catching it!

components/UnitIntroduction.vue
streitlua: I understand that it might be cleaner to put this text in the manifests, but I don't know if it would be more generic

Issue #155, opened 2021-10-18
Add a button to Luã's Ads Prices extension on https://experiences.hestialabs.org


foucault-dumas:
The user should be able to reach the Ads Prices experiment as easily as he/she reaches the other experiences. This button could link to the https://github.com/hestiaAI/my-worth-extension/blob/main/README.md content but hosted on a https://experiences.hestialabs.org/Ads-Prices page. 
![Capture d’écran 2021-10-18 à 16 23 22](https://user-images.githubusercontent.com/81250365/137751162-1b2c9eec-091e-484c-889f-f81fb0561ac7.png)

valentinoli:
@foucault-dumas Instead of displaying the README in the app I would suggest just linking to it. So the user would click the card and be navigated to the README on GitHub

@streitlua any thoughts?

foucault-dumas:
> @foucault-dumas Let me know if I understand this correctly:
> 
> You want a `/ads-prices` page in the app to display the content of the README for the My Worth extension. Would it not be better to let the page be `/my-worth` and the title "My Worth" instead of "Ads Prices" and the subtitle "How is your attention monetized?" for example instead of simply "Data Experience"
> 
> What do you think?

Sorry I wasn't clear: I think we should change the name "My Worh" for "Ads Prices" everywhere. That's why I proposed /ads-prices (I submit this proposal to the vote in The Eyeballs Signal group and come back here with the result)
For the subtitle, yours "How is your attention monetized?" is better than "Data experience" indeed

PR #186, opened 2021-11-02
feat: Added experience card point to the extension repository

streitlua:
Addressed #155


ffsinger (approved):
1. Add it to `config/workshop.json` too
2. It may be better to open the link in a new tab

Good job !

Issue #188, opened 2021-11-03
Allow to configure whether to show generic data viewer for some experience


streitlua:
Add a flag to the manifest allowing to choose whether we want the generic data viewer component to appear on some specific experience

PR #189, opened 2021-11-03
feat: allow configuring whether to show data explorer for each experience

streitlua:
Addressing #188.

Added a key `showDataExplorer` to the manifest of each experience, controlling whether the data explorer component is shown at the bottom of the page or not.
The default value is true.


ffsinger (approved):
All good

PR #190, opened 2021-11-03
feat: each experience can have a different custom pipeline

ffsinger:
This will be necessary because we are soon going to replace twitter experiences, using custom pipelines instead of sparql pipelines


andreaskundig (approved):
thanks

Issue #202, opened 2021-11-04
bug: file tree for generic data viewer is inconsistent


streitlua:
We use unzip from unzipit.js to extract content from a zip file.
To build a file tree for the generic data viewer, we rely on the fact that unzipit.js creates an entry for each file inside the zip, including folders.
However, this is not always the case, so we might have to create the file tree structure from the name of the files only.

valentinoli:
@streitlua 

Sorry, I couldn't handle myself. Might this work? Try running it in the browser console

```js
const files = ['a/b/c.json', 'a/b.json', 'a.json', 'a/b/d.json']
const fileParts = files.map(f => f.split('/'))

const result = []

fileParts.forEach(arr => {
  cur = result
  arr.forEach((value, idx) => {
      if (idx + 1 === arr.length) {
          // leaf node
          cur.push({ value, children: [] })
      } else {
          const node = cur.find(n => n.value === value)
          if (!node) {
              const n = { value, children: [] }
              cur.push(n)
              cur = n.children
          } else {
              cur = node.children
          }
      }
  })
})
```

Output:

```js
[
    {
        "value": "a",
        "children": [
            {
                "value": "b",
                "children": [
                    {
                        "value": "c.json",
                        "children": []
                    },
                    {
                        "value": "d.json",
                        "children": []
                    }
                ]
            },
            {
                "value": "b.json",
                "children": []
            }
        ]
    },
    {
        "value": "a.json",
        "children": []
    }
]
```

streitlua:
@valentinoli thanks for this! I am refactoring the code so that the whole file extraction pipeline is easier to modify and maintain :)
This way, I'll separate: 1. the extraction of all files from zips and inner zips; 2. the creation of a file tree structure from the list of files; 3. the creation of items suitable for `v-treeview`.

valentinoli:
Sounds good! Good luck

Issue #203, opened 2021-11-05
refactor: unify file extraction logic accross all use cases


streitlua:
It would be faster to handle file loading and extraction in the same way for the normal pipeline, for the File Explorer, and for the Generic Data Viewer, since now we load the files and do the zip extraction multiple times in an unnecessary way.

PR #204, opened 2021-11-05
Fixed file tree in UnitFileViewer

streitlua:
Fixes bug #202 
Makes part of the refactoring needed for #203 

The logic of file extraction and building of a file tree has been moved from `components/UnitFileViewer.vue` to `utils/file.js`.
Created a class that could maybe handle everything file-related in the future.


ffsinger (approved):
I don't know well this part of the code but it looks good to me

PR #207, opened 2021-11-10
refactor: Unified file extraction logic for all pipelines and use cases

streitlua:
Addressing #203 

Changes:
- moved all the file extraction logic to the class `FileManager`
- now `UnitFiles.vue` only manages Uppy and returns all the files that were input without any processing
- the real file extraction and preprocessing logic happens in `TheDataExperience.vue` and wherever some place of the code asks the File Manager for data
- allowed each file of an experience to have a different file preprocessor
- removed buttons to use cache and clear cache as they complicated the logic (we can bring them back if you think we should)

What this means:
- a File Manager object is created once in `TheDataExperience.vue`
- every new custom pipeline must take a File Manager object as an argument
- each experience that preprocesses files needs the property `preprocessors` in its manifest
- the manifest also needs to have a `files` field with all the file paths that are used in the rdf pipeline

Tested experiences ✅:
- Uber
- Tinder
- Twitter
- Data Explorer
- Tracker control

Did not test ❌:
- Facebook

Don't hesitate to give me feedback!


andreaskundig:
I tried with my own facebook file and I got the  page "An error occured" and in the console
```
Error: ads_information/advertisers_you've_interacted_with.json doesn't exist in the file manager
```
I prefer the behavior of the current version, where I can see all the visualizations except Ad interactions which gives me a *No relevant data found* message when I run it.

ffsinger:
Preliminary tests:

- There seems to be a big performance improvement for twitter, very nice.
- The message "Successfully processed in x sec." is a bit counter-intuitive because it doesn't take into account the pipeline processing time (e.g. for twitter it says 0.385 sec. even though the pipeline takes 1min to run). But we discussed that we may remove the auto-run for custom pipelines so I don't know what we should do.
- As Andreas said, facebook crashes when a file is missing. It is essential to deal with this when `"allowMissingFiles": true` is set in the experience config.

streitlua:
Thanks for your comments :)

- Indeed the elapsed processing time is not correct, I'm trying to understand why.
- I was not aware of this behaviour for facebook (I had never tested the facebook experience before since it doesn't have test data on the app), I will try to fix it.
I don't know how hard a fix will be, maybe we can discuss this on the meeting tomorrow morning and we decide then if we merge the PR with the bug or if we try to fix it.

ffsinger (approved):
All good for the code, nice job !

ffsinger:
Note that the crash occurs in any experience where a file is missing. Even if `allowMissingFiles` is set to false, we should handle a missing file by showing a message to the user without crashing the app. See how it was done before: 

Screenshot_20211110_172332

streitlua:
I believe to have addressed the issues that you mentioned, could you please test the PR again? Thanks!

andreaskundig (approved):
The bug is fixed, it works as before with my facebook data.

PR #208, opened 2021-11-11
feat: Add new Sunburst build with d3

fquellec:
A new component to build interactive sunbursts. Replaced the Twitter old ones and add a parameter in the manifest to specify graphs ids.


manifests/experiences/twitter/twitter.json
```diff
@@ -41,7 +41,8 @@
     },
     {
       "customPipeline": "targetingTree",
-      "visualizations": ["sunburst-targeting"],
+      "visualizations": ["Sunburst.vue"],
+      "graphId": "tcTop",
```
ffsinger, line 45: It's not completely coherent to have the graphId like this because the block may contain multiple visualizations. Possible solutions:
- Generate a random ID inside Sunburst.vue
- Decide that 1 block should not contain more than 1 viz
- Leave it like this until it's a real problem

ffsinger (changes requested):
1. I'm getting a lot of the following in the console: 

```
Uncaught TypeError: d.data.name is undefined
    mouseover Sunburst.vue:288
    [...]
```

2. You should update the text description of the block (it's outdated), and explain that you can click on the segments to zoom in (it took me a minute to notice this).

streitlua:
The new graph is very nice to use and seems user-friendly as well :)
I agree with what Florian said about the id of the viz (why do you actually need to specify the id in the manifest?).
Did not have any error messages on mouseover though.

manifests/experiences/twitter/twitter.json
```diff
@@ -41,7 +41,8 @@
     },
     {
       "customPipeline": "targetingTree",
-      "visualizations": ["sunburst-targeting"],
+      "visualizations": ["Sunburst.vue"],
```
streitlua, line 44: I don't know enough about how this is done but why keep the `.vue`?

andreaskundig: The ".vue" extension tells us what kind of visualization to use. If we were more consistent, the others would have a ".vega" extension

ffsinger: In UnitQuery we are differentiating between Vega and D3 graphs by doing `this.vizNames.filter(n => n.endsWith('.vue'))`. Could be done differently but I don't see a reason to change it now, let's wait until all vega graphs are replaced by D3

manifests/experiences/twitter/twitter.json
andreaskundig: I'm in favor of generating the Id inside of Sunburst.vue, it doesn't need to be random, it could use a counter or something.

Element ids are supposed to be unique on the page, and it's possible that we end up having two instances of the same experience next to each other on the same page.

fquellec: Thanks for the review, I just find out that each component has a unique ID which can be accessed as this._uid, I will use this ID instead

fquellec:
> 1. I'm getting a lot of the following in the console:
> 
> ```
> Uncaught TypeError: d.data.name is undefined
>     mouseover Sunburst.vue:288
>     [...]
> ```
> 
> 1. You should update the text description of the block (it's outdated), and explain that you can click on the segments to zoom in (it took me a minute to notice this).

I cannot reproduce the issue, let's leave it here for now

ffsinger:
I investigated more carefully and the problem is actually only there:

Screenshot_20211111_171238

One of the targeting values is 'undefined'

ffsinger (approved):
Actually the problem was already there with the previous sunburst, so it's more likely to be related to the pipeline. Sorry for this, I'll create a separate issue and merge this PR.

Good job !

PR #210, opened 2021-11-12
feat: put back run buttons for custom pipelines

ffsinger:
I didn't manage to show the progress circle on the button like for sparql. It would be nice if someone knows why

Note: I removed this part

```
computed: {
    disabled() {
      return !this.fileManager.preprocessedTexts
    }
  }
```

for two reasons:
- It's always false because preprocessedTexts is initialized as an empty object
- Even if the value changed, I'm not sure if vue would detect it (see [reactivity in depth](https://vuejs.org/v2/guide/reactivity.html))

### DIFF
--- a/components/UnitCustomPipeline.vue
+++ b/components/UnitCustomPipeline.vue
@@ -32,8 +32,11 @@
 <script>
 import { processError } from '@/utils/utils'
 import FileManager from '~/utils/file-manager'
+import BaseButton from '~/components/BaseButton'
 
 export default {
+  name: 'UnitCustomPipeline',
+  components: { BaseButton },
   props: {
     fileManager: {
       type: FileManager,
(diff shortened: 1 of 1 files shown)


streitlua:
I think that the issue with the progress icon is related to using web workers: it seems that the progress icon doesn't have time to be updated before the code blocks to wait for the result of the custom pipeline.

ffsinger:
> I think that the issue with the progress icon is related to using web workers

I'm not using web workers for the custom pipeline (I wanted to but couldn't find a way).

> it seems that the progress icon doesn't have time to be updated before the code blocks to wait for the result of the custom pipeline.

This I agree, but I don't really know why it works fine for all the other buttons and not in this case.

components/UnitCustomPipeline.vue
```diff
@@ -59,25 +62,27 @@ export default {
...
-      } catch (error) {
-        console.error(error)
-        this.error = true
-        this.message = processError(error)
-        this.$emit('update', { error })
-      } finally {
-        this.status = true
-        this.progress = false
-      }
+      setTimeout(async () => {
```
streitlua, line 69: I found a solution: set the heavy computation to start after 1 ms using setTimeout, this allows the dom to be updated.

ffsinger: Nice, thank you !

valentinoli: Sounds like a use case for [$nextTick()](https://vuejs.org/v2/api/#Vue-nextTick)

```
await this.$nextTick()
```

PR #214, opened 2021-11-15
feat: Kepler integration

andreaskundig:
Adds the experience uber-kepler, which includes a kepler visualization. 

Kepler runs as the static page /kepler which is included on the experience page through UnitIframe. It would probably be possible to include it directly inside a vue component that we write ourselves and imports react. Since the visualization only receives data, we can do something simple by hand without an integration library like vuera.

In order to let kepler process the csv input with its own processor, everything that works with results has been refactored to work with an untyped variable instead of something of the form {header, results}

Visualizations are now responsible for checking themselves that they can display the data passed to them. If the data is not in the right format, they should display an error message. If an experience shows such an error message, it means you need to reconfigure it in order to have data in the correct format. The import page will show an error for any data that cannot be displayed in a table.

This pull request only makes the integration work. There remains a lot of work to configure kepler properly. Meanwhile, we can hide it from the user through configuration. The pull request should be merged when we're sure it doesn't break existing experiences.


pages/import.vue
```diff
@@ -67,37 +67,44 @@
...
         </v-row>
         <v-row
-          v-for="(graphName, index) in allVueGraphNames[i]"
-          :key="`viz-${i}-${index}`"
+          v-for="(graphName, vizVueIndex) in allVueGraphNames[resultIndex]"
+          :key="`viz-${resultIndex}-${vizVueIndex}`"
         >
           <v-col>
-            <vue-graph-by-name :graph-name="graphName" :values="allItems[i]" />
+            <vue-graph-by-name :graph-name="graphName" :data="result.result" />
```
ffsinger, line 94: Let's be careful with this kind of change because it breaks the import for zips that were generated before this PR. Fortunately we haven't used the import feature in production yet.

static/kepler/kepler-init.js
```diff
@@ -0,0 +1,89 @@
+/* global React, ReactDOM, ReactRedux, Redux, KeplerGl */
+/* eslint no-undef: "error" */
+// based on the example at https://github.com/keplergl/kepler.gl/blob/master/examples/umd-client/index.html
+
+const MAPBOX_TOKEN =
```
ffsinger, line 5: I'm getting this error when running the kepler experience : 

`
Cross-Origin Request Blocked: The Same Origin Policy disallows reading the remote resource at https://events.mapbox.com/events/v2?access_token=pk.eyJ1IjoiYW5kcmVhc2t1bmRpZyIsImEiOiJja3ZxcnlmNXc2ZzUwMnFva2F2a3Q1azU5In0.NrvCU8OKlkwJOVFOgZzTzA. (Reason: CORS request did not succeed)
`

It still seems to work though

ffsinger (approved):
Ok for me.

It will create conficts with PR #217 but let's merge Kepler first.

streitlua (approved):
The pipeline changes look good for me, it doesn't seem that any experience has been broken.
The iframe for Kepler doesn't appear correctly (wrong size and cut out sometimes), but this can be fixed later.
Also, it seems that the data displayed in the map doesn't match the input data (there's a point in Algeria and in Spain but the data is only in the UK), maybe this is related to the CORS requests that fails (pointed out by @ffsinger).

streitlua (changes requested):
Apparently, the Twitter experience is broken for the advertiser search. 
@ffsinger is working on it

streitlua (approved):
The Twitter search advertiser component seems to work now. (although you can only make one search, the following searches have no effect - apparently this has been fixed in #217)

ffsinger:
> apparently this has been fixed in #217

Sorry, I actually haven't been able to fix it properly, so I opened a new issue (#219)

Issue #215, opened 2021-11-15
Add experiences "available on demand" on https://experiences.hestialabs.org


foucault-dumas:
Please add experiments under construction or to come (Apple, amazon, uber, spotify, google... visible on test.hestialabs.org) to the experiences.hestialabs.org page. It must be clear that they are not clickable: Text "available on demand" + greyed out ?

streitlua:
I'm working on it!
Would it be ok to have the text "Under construction..." instead? Would we really be able to answer any demands?

pdehaye:
Why not "Contact Us"?

foucault-dumas:
Yes, why not, but it means:

### Public experiences
FB / TW / AdRadar / Tinder (clickable)

### Other experiences ([contact us](mailto:contact@hestialabs.org))
Google / Apple / Amazon / Spotify / Uber / Netflix... (not clickable)

streitlua:
Did you have something like this in mind? (we won't show these experiences exactly, this is just for testing purposes)
On the main page:
![image](https://user-images.githubusercontent.com/16099301/141959429-cdf8cdd5-21cb-4910-baa7-63706c5539b9.png)
![image](https://user-images.githubusercontent.com/16099301/141959464-21601c34-cdcd-4662-8898-da533cfc32e5.png)

In the side bar:
![image](https://user-images.githubusercontent.com/16099301/141959533-385f0a86-bd98-4064-bdc7-e379b07ca043.png)
![image](https://user-images.githubusercontent.com/16099301/141959585-795e5b3a-e776-408b-8c41-f2125c529740.png)

foucault-dumas:
Yes, but, really, I would prefer "Other experiences" and to not display "not ready"
While you're here, would you mind changing "the newsletter of HestiaLabs" --> "HestiaLabs' newsletter" ?

streitlua:
I'm still not sure I understand the goal of showing disabled experiences.

After discussing this with the team this morning, we found that "public experiences" vs "other experiences" can be confusing for users, will any experience actually be private?
Wouldn't "discover our experiences" vs "future experiences" be a better distinction?

What text would you like to show in the disabled experiences?

I will change the text for the newsletter.

foucault-dumas:
[...]

PR #216, opened 2021-11-16
feat: separated data experiences and refactored code

streitlua:
Following the discussion on #215, we separated the experiences into "public" and "available on-demand".
A new experience "other" has been added, with a link to contact us.

The logic of the data experience list was refactored, so we had to make changes to `index.vue` and `default.vue` as well.


ffsinger (approved):
Code looks good to me. Is it on purpose that the title "Understand your data" disappeared ? It gave the user an indication of what this page is about

streitlua:
> Code looks good to me. Is it on purpose that the title "Understand your data" disappeared ? It gave the user an indication of what this page is about

Yes it is on purpose, it's part of the discussion in #215, although I agree that "public experiences" could sound weird for an user

streitlua:
> code in [918e916](https://github.com/hestiaAI/hestialabs-experiences/commit/918e916a93339b4166c8c8689ec2a7a9fdb03615) seems a bit repetitive to me, would be better to have a loop in there

I agree that the code was repetitive at that point, I hope that you find it better now!

> Also, it is generally not a best practice to use `<br>` to create margins and better to use CSS.
> 
> Vuetify offers classes that are a better way to achieve this, `mb-` classes for bottom margins.

I will keep that in mind and try to replace the `<br>` by mb- classes :)

Issue #194, opened 2021-11-03
Move if-else logic inside unit-query


andreaskundig:
There is conditional logic inside UnitQuery and also outside of it in [TheDataExperienceDefault](https://github.com/hestiaAI/hestialabs-experiences/blob/ae782b1043e29dca0006f5d044b7e24de9c55b96/components/TheDataExperienceDefault.vue#L31)

Move it all inside the component

```
<unit-query
            v-if="defaultViewElements.customPipeline"
            v-bind="{
              visualizations,
              defaultViewElements,
              selectedExample,
              customPipeline:
                customPipelines[defaultViewElements.customPipeline],
              inputFiles,
              i: index
            }"
            @update="onQueryUpdate"
          />
          <unit-query
            v-else
            v-bind="{
              selectedExample,
              visualizations,
              defaultViewElements,
              query: queries[index],
              i: index
            }"
            @update="onQueryUpdate"
          />
```

PR #217, opened 2021-11-16
feat: SQL pipeline

### LINKED ISSUE #147: Configurable pipeline: input >... > query results
# The shortcut we have now

Currently we have a flag **queryShortcut** that's configured in the experience, for example in **manifests/experiences/tracker-control/tracker-control.json**

When the flag is true, the usual pipeline: 
```
input file -> rdf -> sparql -> query results
```
Is replaced with:
```
input file -> custom code (UnitSql) -> query results
```
The custom code only parses a csv input file, converts it to json, and uses that json as query results.

# We want a configurable pipeline

Let's stop call it a shortcut, and call "pipeline" the part that brings you from the input to the query results.

What we want is to able to configure alternative pipelines. Either:
- sparql pipeline
- sql pipeline
- custom pipeline

The sql pipeline does not exist yet, it's going to do something similar as the sparql/comunica pipeline but with sql/db.js. Both of these pipelines would be part of the application code, and would only be referred by name in the manifest. VueGraphByName.vue is an example of how this could work, maybe you find a way to do it without having to hardcode a switch.

The custom pipeline would be a javascript file in the manifest directory. An example would be the way we load d3 code in UnitD3Viz.vue.

Every pipeline should validate the format of it's input files #145

ffsinger:
Implement a SQL pipeline and convert Twitter experiences to this pipeline. Also includes some refactoring for the parametrized experiences.

Note: some visualizations (e.g. sunburst) look a bit different, because I used this opportunity to include engagements (before we were using only impressions).

Close #147


ffsinger:
PR #214 should be merged before

ffsinger:
Kepler has been merged and conflicts have been solved, so this PR is now ready for reviews

ffsinger:
This PR will close #194

PR #221, opened 2021-11-17
fix: #218 and various

### LINKED ISSUE #218: All experiences are accessible through URL
Example: https://experiences.hestialabs.org/twitter-advertisers should not exist on this deployement. Is there a way to take them out of the build ?

We can also access "unavailable" experiences such as https://experiences.hestialabs.org/apple.

valentinoli:
* fixed #218, added a navigation guard middleware to `pages/_key.vue`
* moved `enabledExperiences`/`disabledExperiences` to the Vuex store
* improved newsletter code, better to keep code in `<script>` and `<template>` as clean as possible
* changed config loading to load the config in `manifests/index.js`
* added experience filter in `manifests/index.js` based on the config

### DIFF
--- a/components/TheAppBar.vue
+++ b/components/TheAppBar.vue
@@ -59,9 +59,7 @@
       </template>
       <div class="mt-6">
         <logo-img width="250" />
-        <the-data-experience-list
-          v-bind="{ enabledExperiences, disabledExperiences, big: false }"
-        />
+        <the-data-experience-list />
       </div>
     </v-navigation-drawer>
   </div>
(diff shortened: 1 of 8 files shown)


ffsinger (approved):
Good !

ffsinger:
Actually, I just pushed a small change: greyed out experiences now also return 404 because they're not supposed to be available

valentinoli:
Ah didn't think of that! Thanks

valentinoli:
better to do

```js
!store.getters.enabledExperiences.find(x => x.key === params.key)
```

valentinoli:
When this is ready I need to merge it into `refactor/components` branch before organizing components into directories

ffsinger:
I think it can be merged into master

PR #226, opened 2021-11-21
refactor(components): enforce style and enhance organization

valentinoli:
This PR is about refactoring Vue components and enforcing conventions and style based on the [Vue style guide](https://vuejs.org/v2/style-guide/). It is also about taking better advantage of the auto-importing feature of the [Nuxt components module](https://github.com/nuxt/components). Here is an overview of the suggested changes in the PR:

* I noticed that someone had added unnecessary import statements and `components` declarations all over the codebase. Please note that these are *not needed* with the [Nuxt components module](https://go.nuxtjs.dev/config-components) that takes care of all that for us! So I removed them.
* I enabled [auto-importing components](https://go.nuxtjs.dev/config-components) within nested directories (without path prefixing) with the following Nuxt config:
```js
components: [
  { path: '~/components', pathPrefix: false, extensions: ['vue'] }
]
```
This means a component name only consists of the file name and not (path prefix + file name). This can be overridden for some directories if we want. The VSCode Vetur extension provides autocompletion for all these components as well as Vuetify components. I also added the following ESLint rule to maintain consistency in the future:

```js
'vue/component-name-in-template-casing': [
  'error',
  'PascalCase',
  {
    registeredComponentsOnly: false,
    // https://github.com/vuejs/eslint-plugin-vue/issues/1722
    ignores: ['/^component|client-only|keep-alive$/']
  }
]
```

Although Vuetify components are suggested as `kebab-case`, the ESLint VSCode extension fixes that when you save the file. I added extensions recommendations to `.vscode/extensions.json` and everyone in the team should have them (at the moment, Vetur, ESLint, and Prettier).

* I suggest we use `PascalCase` instead of `kebab-case` in component names. See style guide: [Component name casing in templates](https://vuejs.org/v2/style-guide/#Component-name-casing-in-templates-strongly-recommended)

* I changed many component names as per these two recommendations from the style guide: [Tightly coupled component names](https://vuejs.org/v2/style-guide/#Tightly-coupled-component-names-strongly-recommended) and [Order of words in component names](https://vuejs.org/v2/style-guide/#Order-of-words-in-component-names-strongly-recommended)

* I added [Vue mixins](https://vuejs.org/v2/guide/mixins.html) where applicable


streitlua (approved):
It looks all good to me!
It was me who imported and registered the components manually because they always appear as unrecognized HMTL tags and I can't navigate to them. This still hasn't changed with the PR but I guess that this is not such a problem :)

valentinoli:
> It was me who imported and registered the components manually because they always appear as unrecognized HMTL tags and I can't navigate to them. This still hasn't changed with the PR but I guess that this is not such a problem :)

That's ok don't worry :)
It must be related to your editor settings. I do not get any warnings or errors. What do you mean by "can't navigate to them"?

andreaskundig:
We discussed this among the reviewers and we're all ok with it. There are some other branches we need to merge first, then we'll see how the merge then looks.

valentinoli:
Ok! Good to hear :)
Yes, we need to wait for other PRs to merge since there are overlaps and changed file names as well. So some manual resolution might be needed.

streitlua:
> What do you mean by "can't navigate to them"?

I mean that I cannot ctrl+click and jump to the corresponding file/code without registering the components, but this is not that important

andreaskundig:
And you should know that Lua is on [Webstorm](https://www.jetbrains.com/webstorm/). I'm glad I'm not the only one not using vs code, I'm still on emacs, but we're probably all using the same lsp server.

valentinoli:
There must be a way to get that working somehow

valentinoli:
> And you should know that Lua is on [Webstorm](https://www.jetbrains.com/webstorm/). I'm glad I'm not the only one not using vs code, I'm still on emacs, but we're probably all using the same lsp server.

Oh, I didn't know :(
I haven't tried other editors for Vue but from what I read VSCode is encouraged for Vue projects. Main extensions for language support are VSCode extensions I think

andreaskundig:
Merge this after the branch [real-generic-date-viewer](https://github.com/hestiaAI/hestialabs-experiences/tree/real-generic-date-viewer)

Issue #191, opened 2021-11-03
A first integration test


andreaskundig:
We need tests to catch regressions. If your new feature breaks an old feature, a test that fails should tell you about it. This first test will not catch many errors, but setting it up will make it easier to write more tests.

Set up a test that runs the application and opens an experience.
Configure github to run the test at every commit, or once a day. See github CI https://docs.github.com/en/actions/automating-builds-and-tests/about-continuous-integration

The application is built with nuxt, and was created with the [create-nuxt-app command line tool](https://github.com/nuxt/create-nuxt-app/blob/master/README.md). 

One way to go about adding tests would be to see if the command line tool can add a testing framework after the fact. If not, youc ould follow the conventions of that tool to install a testing framework by hand. On the other hand, maybe there's another good way of adding a test framework to a nuxt app.

There are several test frameworks to choose from. I worked with Jest in react, it seems it's also a popular option for vue. Maybe do a quick search to see if there's anything that overwhelmingly better.

valentinoli:
Have you considered End-2-end testing? Like Cypress: https://www.cypress.io/
I have not any experience with such frameworks - is this something you're looking for?

andreaskundig:
I've used selenium tests before and they're a real pain. It looks like Cypress is something like it, it also runs in the browser, and that it works better, so why not try it out. Apparently they have a github ci integration too, but does this mean github ci runs a browser? 

Apparently [people ](https://stackoverflow.com/questions/66217682/should-i-use-both-cypress-and-jest-together) are using it with jest and testing-library which are a bit lighter since they don't need a browser and can be used for unit tests. They would be sufficient for a simple integration test, so if we start with that we have something in place that also helps us write more fine grained tests.

The simpler frameworks like jest run without browser so there are less parts that can fail. Selenium was a pain in part because it depends on the browser and new browser versions meant that new drivers needed to be installed constantly, and tended to break tests. I don't know how cypress deals with this, probably better, but I think there is less risk using a simpler tool at first. The simpler tool is also likely to run faster. Once we have that, we could try cypress too, it does look pretty cool.

andreaskundig:
We have the two first tests

The only thing missing now is a test that loads the whole application. This would be the first integration test.

andreaskundig:
I would expect that there's something in between e2e running in a browser, and component tests for every single component. 

[...]

PR #232, opened 2021-11-23
Framework for writing tests

ffsinger:
This PR aims at setting up a way of writing and running tests easily. It consists of the following additions:

- Add [Jest](https://jestjs.io/) for running test suites with `npm run test`.
- Unit tests on `.js` files can be written using Jest. An example is available in [utils/csv.test.js](https://github.com/hestiaAI/hestialabs-experiences/blob/test/unit-tests/utils/csv.test.js)
- Unit tests on components can be written with the help of Jest and [Vue test utils](https://vue-test-utils.vuejs.org/). An example is available in [components/\_\_tests\_\_/UnitFilterableTable.test.js](https://github.com/hestiaAI/hestialabs-experiences/blob/test/unit-tests/components/__tests__/UnitFilterableTable.test.js). It shows how to mount a component, mock the vuex store, and find specific HTML elements.

Notes:
- We could have used [testing-library](https://testing-library.com/docs/vue-testing-library/intro), which is an abstraction on top of vue-test-utils. But I found it to be less powerful than vue-test-utils.
- You may encounter import errors when writing new tests. That's because Jest does not support ES6 syntax. For instance, I had problems with components using D3. The solutions [here](https://stackoverflow.com/questions/60714101/how-to-setup-jest-with-node-modules-that-use-es6) may solve the problem in some cases, but for D3 it didn't help.

TODO:
- Snapshot testing
- E2E testing with [cypress](https://www.cypress.io/) (see #191)

### DIFF
--- a/babel.config.json
+++ b/babel.config.json
@@ -1,0 +1,5 @@
+{
+  "presets": [
+    "@babel/preset-env"
+  ]
+}
--- a/components/UnitFilterableTable.vue
+++ b/components/UnitFilterableTable.vue
@@ -8,6 +8,7 @@
         v-bind="{ headers: tableHeaders, items, search }"
         :hide-default-footer="disabled"
         multi-sort
+        data-testid="data-table"
       >
         <template #item.url="{ value }">
           <a target="_blank" rel="noreferrer noopener" :href="value"> Link </a>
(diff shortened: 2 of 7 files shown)


ffsinger:
The diff of package-lock.json is awfully large, I wonder what happened

PR #239, opened 2021-11-23
fix: better handling of cases where no data is found

ffsinger:
There were errors on the twitter experience when the user inputs an advertiser that doesn't exist in the data. Now the page shows "No relevant data found".

### DIFF
--- a/components/unit/UnitQuery.vue
+++ b/components/unit/UnitQuery.vue
@@ -77,7 +77,7 @@
           </VCol>
         </VRow>
         <template v-if="finished">
-          <template v-if="result">
+          <template v-if="hasData">
             <VRow>
               <VCol
                 v-for="(specFile, vegaIndex) in vegaFiles"
@@ -115,7 +115,7 @@
           </template>
           <template v-else>
             <VRow>
-              <VCol>
+              <VCol style="text-align: center">
                 <p><i>No relevant data found</i></p>
               </VCol>
             </VRow>
@@ -194,6 +194,9 @@ export default {
     },
     vegaFiles() {
       return this.vizNames.map(n => this.vega[n]).filter(n => n)
+    },
+    hasData() {
+      return this.result && (this.result.headers?.length ?? 1) > 0
     }
   },
   methods: {
--- a/components/unit/UnitQuery.vue
+++ b/components/unit/UnitQuery.vue
@@ -25,7 +25,7 @@
             v-for="(specFile, vegaIndex) in vegaFiles"
             :key="vegaIndex"
             cols="12"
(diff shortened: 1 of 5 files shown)


streitlua (approved):
Looks good to me, although the problem of the results not refreshing when we change the query still exists.

ffsinger:
> the problem of the results not refreshing when we change the query still exists.

It has been fixed in #234 and when I test it works. Are you sure ? How does it happen ?

components/unit/UnitQuery.vue
```diff
@@ -115,7 +115,7 @@
           </template>
           <template v-else>
             <VRow>
-              <VCol>
+              <VCol style="text-align: center">
```
andreaskundig, line 118: It's better if we can avoid using inline styles, they make it harder to have a consistent look everywhere. In this case, you should be able to use [justify from the vuetify grid system](https://vuetifyjs.com/en/components/grids/#justify)

andreaskundig (changes requested):
Otherwise I trust Luà's assessment that everything's ok

streitlua:
I found another issue when testing with the `twitter.zip` archive (although this is likely related to the sql pipeline and not this PR):
image
image

streitlua:
> > the problem of the results not refreshing when we change the query still exists.
> 
> It has been fixed in #234 and when I test it works. Are you sure ? How does it happen ?

You are right, it is working. I was testing random words and then `apple` but was still seeing no results. The search is case-sensitive (maybe it shouldn't be?) so I can see results when searching `Apple`.

components/unit/UnitQuery.vue
ffsinger: If I use something like this:

```
<VRow justify="center">
  <VCol cols="4">
    <p><i>No relevant data found</i></p>
  </VCol>
</VRow>
```

It centers the column but not the text inside it:

Screenshot_20211124_094415

I would say that it doesn't look as good as having the text fully centered, what do you think ? Can we achieve this without using `style="text-align: center"` ?

andreaskundig: It looks fine to me. I don't like centered text. 

And the text in the description is also left-justified (could be something I did, actually).

andreaskundig: text-center is also an acceptable solution

ffsinger:
It would indeed be better if it wasn't case-sensitive, I'll see if I can fix this

ffsinger:
Case-sensitivity has been fixed in commit 557b004a13732b86bd5e532405b96e468fcedc68

Issue #237, opened 2021-11-23
Make the data export box more flexible


foucault-dumas:
### Long term:
As a collective manager, I want to be able to define the content of the export box of an experiment and the associated subset of extracted data according to a defined objective (in agreement with possible partners)

### Short term:
As The Eyeballs, I want the MEPs participating in the Tracking-free ads coalition Twitter experience to be able to...:

1. ...indicate their name and email or not (optional)
2. ...opt-in to send us the advertisers who targeted them and the criteria they used to do so
3. ...be informed with this text: "I consent to sharing my targeting criteria for the exclusive purpose of producing screenshots highlighting how I have been targeted by Facebook, Google or other actors of interest to the [Tracking-Free Ads Coalition](https://trackingfreeads.eu). These screenshots will then be shared with the Tracking-Free Ads Coalition. The raw data will be deleted within 2 months. For further information please refer to our [privacy policy](https://hestialabs.org/en/legal/)."

This new experience can be set up either in a new card (probably ephemeral) in experiences.hestialabs.org, or on a dedicated site tfac.hestialabs.org, or in a new site campaigns.eyeballs.hestialabs.org -> this choice depends

1. on the potential speed of implementation for this tracking-free ads campaign
2. on its relevance with the achievement of the above long term objective

(I don't think I have the elements alone to make the right choice but I'll gladly discuss it with you)

ffsinger:
Closed by #248

PR #240, opened 2021-11-23
feat: consent form upgrade

ffsinger:
Added support for text input fields in the consent form. Also removed `saveFiles` parameter that is no longer usable since the `fileManager` refactoring.

### DIFF
--- a/components/unit/consent-form/UnitConsentForm.vue
+++ b/components/unit/consent-form/UnitConsentForm.vue
@@ -88,7 +88,7 @@ export default {
   watch: {
     includedResults: {
       deep: true,
-      handler(includedResults) {
+      handler() {
         this.resetStatus()
       }
     }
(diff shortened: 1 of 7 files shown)


ffsinger:
Address part of #237

streitlua (approved):
Looks good, seems to be working for me!

I suggest adding a consent section to one of the configs as an example.
Also, what is the deal with the `readonly` prop? It is never given so it defaults to false all the time.

ffsinger:
> I suggest adding an input consent section to one of the configs as an example.

It has been added as an example to `config/config.json`, and it's documented in the readme.

> what is the deal with the readonly prop? It is never given so it defaults to false all the time.

It is used in `pages/import.vue`, line 64. When we import a zip, the consent form is read-only

streitlua:
My bad, it's all good to me then!

PR #242, opened 2021-11-24
feat: Uber overview v2

fquellec:
Some updates on the uber experience, mainly the overview graph made clearer.


streitlua (approved):
It would be nice to make all the days of the week appear on the lower middle plot, but otherwise the viz is very nice!

PR #243, opened 2021-11-24
feat: Tracking-Free Ads Coalition consent form

ffsinger:
Setup of the form as described in  #237


ffsinger:
I'm merging it directly as it shouldn't cause any problem

Issue #241, opened 2021-11-24
Facebook experience broken


fquellec:
When I try to upload my zip file (3.3 GB), I get this error in the upload module:
![image](https://user-images.githubusercontent.com/17878016/143207450-d074dcf3-756b-450b-8c3a-5a20517a4507.png)

When I use the Charles archive (200Mb), I pass the upload but I get this error when running the first cell:
![image](https://user-images.githubusercontent.com/17878016/143218721-f545697c-8241-4c33-8f03-25b4705b6cd2.png)

Using Annaelle's archive (20Mb) everything works.

The archives are available on the kDrive under Projects/TheEyeballs

streitlua:
It seems that this is related to the extraction of the files using the library `unzipit.js`.
I found the [code where the error is thrown](https://github.com/greggman/unzipit/blob/master/src/unzipit.js#L201) and I think that the problem is either related to a malformed zip archive, or to the large size of the given archive.

streitlua:
I tested the Facebook experience with Charle's and Anaelle's archives and everything worked correctly.
Could you check if your zip archive is formatted correctly? Maybe try extracting everything from it and creating a new one.

streitlua:
Actually, there is indeed a problem on the first viz with Charle's archive, apparently the data validator fails. I'm investigating.

streitlua:
There shouldn't be any more issues with the facebook experience (other than the fact that it's currently running on the sparql pipeline which is very slow)

PR #244, opened 2021-11-25
fix: unitvegaviz and chartview dont crash when theres no data

streitlua:
Addressing #241 

The problem is related to how `UnitVegaViz` and `ChartView` handle the given data when it's malformed or empty.
Changes:
- the error message now depends on whether the data is in the correct shape or if it is empty
- removed references to the vega graph when it doesn't exist in `UnitVegaViz`

Please give feedback if you see another way to solve this issue!

### DIFF
--- a/components/unit/UnitVegaViz.vue
+++ b/components/unit/UnitVegaViz.vue
@@ -110,9 +110,11 @@ export default {
       this.draw()
     }
   },
-  async mounted() {
-    if (this.$refs.graph) this.width = this.$refs.graph.offsetWidth
-    if (this.isValid && !this.isEmpty) await this.draw()
+  mounted() {
+    this.width = this.$refs.graph?.offsetWidth ?? 0
+    if (this.isValid && !this.isEmpty) {
+      this.draw()
+    }
   },
   methods: {
     async draw() {
--- a/components/unit/UnitVegaViz.vue
+++ b/components/unit/UnitVegaViz.vue
@@ -1,6 +1,6 @@
 <template>
   <div>
-    <template v-if="isValid">
+    <template v-if="isValid && !isEmpty">
       <div :id="divId" ref="graph"></div>
       <VRow>
         <VCol cols="6 mx-auto">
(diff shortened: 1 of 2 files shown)


components/unit/UnitVegaViz.vue
```diff
@@ -98,13 +106,13 @@ export default {
...
-    specWithItems(v) {
+    specWithItems() {
       this.draw()
     }
   },
   async mounted() {
-    this.width = this.$refs.graph.offsetWidth
-    await this.draw()
+    if (this.$refs.graph) this.width = this.$refs.graph.offsetWidth
+    if (this.isValid && !this.isEmpty) await this.draw()
```
streitlua, line 115: This is the part that seems the most hacky to me, what do you think?

andreaskundig: It looks ok to me. For the style, I would prefer to always use if with braces. I'm suprised that our over-pedantic linter doesn't enforce that. Also width can be 0, so I would write:
```

this.width = this.$refs.graph?.offsetWidth || 0
if (this.isValid && !this.isEmpty) {
   // we don't need to wait, actually
   this.draw() 
}
```

components/unit/UnitVegaViz.vue
```diff
@@ -13,22 +13,27 @@
...
-  return !data || !!(data.items?.length > 0)
+  return (
+    _.every(
+      ['items', 'headers'],
+      field => _.has(data, field) && Array.isArray(data[field])
+    ) &&
+    data.headers.length > 0 &&
+    _.every(data.items, i => _.every(data.headers, h => _.has(i, h)))
+  )
+}
```
andreaskundig, line 33: Another style comment: Multiline boolean expressions are harder to read than if statements. I would try to do something like this

```
const {headers, items} = data
if (!Array.isArray(headers) && !Array.isArray(items) ) {
  return false
}
if (headers.length === 0) {
  return false
}
const missingHeader = items.find(item => item.length < headers.length)
if(missingHeader){
  return false
}
```
(Also I try to avoid lodash if the standard library can do the job without too many complications)

andreaskundig (approved):

components/unit/UnitVegaViz.vue
streitlua: I don't necessarily agree that multiline boolean expressions are harder to read.
Also, I think that lodash's functional and compact style allows writing code that is less error-prone.

fquellec (approved):
looks good to me :)

PR #247, opened 2021-11-25
feat: Search sunburst

fquellec:
New component for the Twitter experience, allowing users to search for an advertiser in a table and display the corresponding sunburst on click.

image

### DIFF
--- a/components/chart/view/ChartViewSearchSunburst.vue
+++ b/components/chart/view/ChartViewSearchSunburst.vue
@@ -1,0 +1,61 @@
+<template>
+  <VContainer>
+    <VRow>
+      <VCol cols="5">
+        <VCard>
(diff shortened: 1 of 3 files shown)


fquellec:
Note: I removed two redundant cards from the Twitter experience (All advertisers and one advertiser)

ffsinger (approved):
Very nice ! The only thing that may be improved is that for now we can deselect an advertiser and the sunburst for this advertiser will stay until we select another one. It would be more intuitive to either: 

1) do not allow deselecting, or 
2) hide the sunburst if no advertiser is selected

ffsinger:
Other small comment: the right arrow in the table on a new line. But it may be a hard fix since it's a vuetify component

PR #248, opened 2021-11-25
feat: various improvements to consent forms for TFAC

ffsinger:
Changes:

- Now we can define a default consent form for a deployement, and override it for a specific experience (either by giving new content or by hiding it completely).
- The section "Results to include" can be hidden. In this case, the data to export is defined through the config.
- Finished setting up the TFAC instance


ffsinger:
I'm merging this now because it's urgent, but feel free to add feedback afterwards

PR #251, opened 2021-11-26
fix: identify sub-experiences with a key

### LINKED ISSUE #250: Identify sub-experiences with a key
Sub-experiences (i.e. the elements of the array `defaultView` in the config) have no identifier. Instead, we are sometimes using the index of their position in the array. While this is fine for some cases, it's not robust in others (e.g. if we are using it in the config like for the attribute `includedResults`) because the position may change.

We should add a key/ID to each sub-experience and use this instead.

ffsinger:
From now on, please add a `key` property to each element of `defaultView`. It can usually be the name of the query/pipeline, but not always.

Maybe we should have a uniqueness check somewhere.

Close #250


streitlua (approved):
The code looks good to me, I don't know how to test it inside the import page but I trust you!

ffsinger:
Thanks, actually we have to merge https://github.com/hestiaAI/hestialabs-experiences/tree/feat/digipower-consent first, and then I'll have to solve some conflicts

ffsinger:
Fixed conflicts, merging now

PR #254, opened 2021-11-26
feat: file viewer for unkown types

andreaskundig:
This covers https://github.com/hestiaAI/digipower-data-experiences/issues/79 in part.

It allows to look at or download files of a type for which we have no specific viewer. It doesn't help in case a specific viewer fails.


ffsinger:
Nice, I'm merging it now so that we have it for Monday.

I'm not sure I understand how it determines whether it shows the content in an IFrame or it downloads the file. Is it necessary to use an IFrame ?

ffsinger (approved):

andreaskundig:
Thank you. 
The whole trick is that the page in the iframe decides if it displays the file or triggers a download.

PR #260, opened 2021-11-29
fix: file explorer does not show an empty file corresponding to the folder itself

streitlua:
For some reason, in some zip archives, the folders are encoded as a file, which messes up the file explorer tree, so we observe empty files like this one:
image
After the fix, this file doesn't appear:
image

### DIFF
--- a/utils/file-manager.js
+++ b/utils/file-manager.js
@@ -283,7 +283,7 @@ export default class FileManager {
   static async extractZips(files) {
     return (
       await Promise.all(
-        files.map(async file => {
+        files.flatMap(async file => {
           if (file.name.endsWith('.zip')) {
             const { entries } = await unzip(file)
             const innerFiles = await Promise.all(
@@ -298,8 +298,10 @@ export default class FileManager {
                 )
             )
             return await this.extractZips(innerFiles)
+          } else if (file.name.endsWith('/')) {
+            return []
           } else {
-            return file
+            return [file]
           }
         })
       )


ffsinger:
So there's no way to detect that these are actually folders ?

streitlua:
> So there's no way to detect that these are actually folders ?

The JavaScript `File` API doesn't allow this, as depending on the OS a folder is not an actual file.
However, the character `/` is not allowed in a folder or file name in any circumstance, so using it to check if the file represents a folder seems robust.
Also, note that this behaviour only happens for some zips that were formatted in a different way.

ffsinger:
I see

I'm wondering if it wouldn't be better to show something instead of hiding the file. Something like "Corrupted folder".

andreaskundig:
If I understand this right, the unnamed folder in the first scrrenshot appears as the file *identifiers.json* in the second one.

streitlua:
> I'm wondering if it wouldn't be better to show something instead of hiding the file. Something like "Corrupted folder".

From what I understand, this weird file corresponds to the folder `MyData`, and we already show the folder in the hierarchy without relying on the file, so I think it's ok to discard files that correspond to folders.

streitlua:
> If I understand this right, the unnamed folder in the first scrrenshot appears as the file _identifiers.json_ in the second one.

I don't think that's right, it's just a file corresponding to the folder `MyData`.

streitlua:
This is what I based my conclusion on:
image
It corresponds to the files in the file manager before the fix, and we can see that the folder `MyData` appears as a file there

andreaskundig:
What happens when you unzip it the usual way on your computer?

streitlua:
> What happens when you unzip it the usual way on your computer?

I can't see any weird file at all, I think that this is a problem due to both how the file was generated, and how `unzipit.js` handles it.

andreaskundig:
Then I don't think we need to show to the user that there's anything unusual.

PR #263, opened 2021-11-29
fix: the download button apply for the additional columns that we add/transform ourselves

fquellec:
Related to https://github.com/hestiaAI/digipower-data-experiences/issues/85
Previously, the download button used the raw csv and not the actual displayed data.

### DIFF
--- a/components/unit/filterable-table/UnitFilterableTable.vue
+++ b/components/unit/filterable-table/UnitFilterableTable.vue
@@ -97,8 +97,9 @@ export default {
         this.error = false
         try {
           const headers = this.headers.map(h => h.text)
+          const filteredItems = this.$refs.tableRef.$children[0].filteredItems
           // Change the items keys to match the headers
-          const itemsWithHeader = this.items.map(i =>
+          const itemsWithHeader = filteredItems.map(i =>
             this.headers.reduce((o, h) => ({ ...o, [h.text]: i[h.value] }), {})
           )
           // update the data
--- a/components/unit/filterable-table/UnitFilterableTable.vue
+++ b/components/unit/filterable-table/UnitFilterableTable.vue
@@ -97,8 +97,12 @@ export default {
         this.error = false
         try {
           const headers = this.headers.map(h => h.text)
+          // Change the items keys to match the headers
+          const itemsWithHeader = this.items.map(i =>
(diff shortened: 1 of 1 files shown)


streitlua:
Everything looks good to me but I agree that it doesn't make much sense to show just the name of one company in the selector.

streitlua:
One thing that seems counter-intuitive to me is that clicking download always downloads the entire data and not just the current filter.
I tested on the generic date viewer with the spotify data and the button shows a spinning progress icon for a very long time (didn't stop spinning after 2 minutes), even if the current selection only has a few entries shown in the table.

fquellec:
> One thing that seems counter-intuitive to me is that clicking download always downloads the entire data and not just the current filter. I tested on the generic date viewer with the spotify data and the button shows a spinning progress icon for a very long time (didn't stop spinning after 2 minutes), even if the current selection only has a few entries shown in the table.

I cannot reproduce the issue, which browser are you using ?

streitlua:
> > One thing that seems counter-intuitive to me is that clicking download always downloads the entire data and not just the current filter. I tested on the generic date viewer with the spotify data and the button shows a spinning progress icon for a very long time (didn't stop spinning after 2 minutes), even if the current selection only has a few entries shown in the table.
> 
> I cannot reproduce the issue, which browser are you using ?

I'm using Firefox. Can you try using the spotify experience with my data and searching for the file `MyData/Playlist1.json` in the date viewer? It only displays two rows, but the download is not immediate as it should, and I guess because all the rows are being downloaded.

fquellec:
> > > One thing that seems counter-intuitive to me is that clicking download always downloads the entire data and not just the current filter. I tested on the generic date viewer with the spotify data and the button shows a spinning progress icon for a very long time (didn't stop spinning after 2 minutes), even if the current selection only has a few entries shown in the table.
> > 
> > 
> > I cannot reproduce the issue, which browser are you using ?
> 
> I'm using Firefox. Can you try using the spotify experience with my data and searching for the file `MyData/Playlist1.json` in the date viewer? It only displays two rows, but the download is not immediate as it should, and I guess because all the rows are being downloaded. Edit: the download finally launched and indeed it contains all the rows and not just the 2.

Thanks, it is fixed now !

PR #264, opened 2021-11-29
feat: Add select to only show a subset of apps in TC experience

fquellec:
Related to https://github.com/hestiaAI/digipower-data-experiences/issues/84
Add select to only show a subset of apps in TC experience, allowing the use of certain applications to be hidden if the user wishes to share their data/viz

### DIFF
--- a/components/chart/view/ChartViewTrackerControl.vue
+++ b/components/chart/view/ChartViewTrackerControl.vue
@@ -1,6 +1,43 @@
 <template>
   <VContainer>
     <VRow>
+      <VCol cols="4" offset="4">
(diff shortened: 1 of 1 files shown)


ffsinger:
Currently we have the name of only one application in the selector (e.g. "Facebook(+11 others)"), and at first it feels like this app is given more importance than the others. Maybe show the name of 3 apps, or none.

streitlua:
Everything looks good to me but I agree with @ffsinger that it doesn't make much sense to show only the name of one company in the selector

fquellec:
fixed

PR #267, opened 2021-11-30
fix: various fixes regarding file upload

ffsinger:
Changes: 

- Fix `file-manager` for multiple files (now it doesn't remove the topmost path element when there are multiple files)
- Fix Uppy restrictions
- Generic experiences are configured to allow multiple files, others are not
- Change information message about file extensions and number of files
- Uppy box is a bit smaller

Related to:
- https://github.com/hestiaAI/digipower-data-experiences/issues/62
- https://github.com/hestiaAI/digipower-data-experiences/issues/59 point 6 within experiences

### DIFF
--- a/utils/file-manager.js
+++ b/utils/file-manager.js
@@ -92,13 +92,13 @@ export default class FileManager {
    * @param {File[]} uppyFiles
    * @returns {Promise<FileManager>}
    */
-  async init(uppyFiles) {
+  async init(uppyFiles, multiple) {
     this.fileList = await FileManager.extractZips(uppyFiles)
     const filePairs = this.fileList.map(f => [f.name, f])
-    if (filePairs.length === 1) {
-      this.fileDict = FileManager.removeTopmostFilenames(filePairs)
-    } else {
+    if (multiple) {
       this.fileDict = Object.fromEntries(filePairs)
+    } else {
+      this.fileDict = FileManager.removeTopmostFilenames(filePairs)
     }
     this.setInitialValues()
     return this
--- a/utils/file-manager.js
+++ b/utils/file-manager.js
@@ -94,9 +94,12 @@ export default class FileManager {
    */
   async init(uppyFiles) {
     this.fileList = await FileManager.extractZips(uppyFiles)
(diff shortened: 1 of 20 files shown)


utils/file-manager.js
```diff
@@ -94,9 +94,12 @@ export default class FileManager {
    */
   async init(uppyFiles) {
     this.fileList = await FileManager.extractZips(uppyFiles)
-    this.fileDict = FileManager.removeTopmostFilenames(
-      this.fileList.map(f => [f.name, f])
-    )
+    const filePairs = this.fileList.map(f => [f.name, f])
+    if (filePairs.length === 1) {
+      this.fileDict = FileManager.removeTopmostFilenames(filePairs)
```
ffsinger, line 99: Maybe we should change how `removeTopmostFilenames` behaves instead, not sure what makes the most sense

streitlua (changes requested):
Nice work ! 
It seems that `this.multiple` is always false, is the prop passed correctly to `UnitFiles`?

About the `removeTopmostFilenames`, it is used for two reasons:
1. if a single non-zip file is given, then the file is renamed to `input.ext` so that we can refer to it without knowing its name
2. if a zip file is given, we remove its name so that we can refer to the files inside without knowing the name of the zip

It's true that currently this is not working for situations when we give many files to uppy, but it seems that your changes broke the reason 2 (try the twitter experience with the sample data and you will see that we can't find the files that we are looking for). Maybe we can just check if `multiple` is true and not do anything in that case?

ffsinger:
> It seems that this.multiple is always false, is the prop passed correctly to UnitFiles

For me it's passed as true when you set `multiple: true` in the config (e.g. `apple.json`)

>  it seems that your changes broke the reason 2

You're right, sorry I didn't notice

streitlua (approved):
Everything looks good now !

streitlua:
> > It seems that this.multiple is always false, is the prop passed correctly to UnitFiles
> 
> For me it's passed as true when you set `multiple: true` in the config (e.g. `apple.json`)

It now works with the latest commit, it might have been an issue related to the application not retrieving the most recent version of the manifest, which happens sometimes :)

PR #273, opened 2021-12-02
feat: generic viz support

ffsinger:
- Refactor visualizations: each block now contains 0 or 1 viz, instead of 0 or more (it wasn't useful).
- Can now define props for visualizations in the json config


fquellec (approved):
All Good ! Thanks again :)

PR #275, opened 2021-12-02
performance: json viewer is now faster and recursively groups elements of a long list together

streitlua:
This is part of the improvement of the performance of the generic file explorer, more specifically the json viewer:
- (https://github.com/hestiaAI/digipower-data-experiences/issues/66)

### What changed?
Right now, we always group elements of a big list into groups of 50. If a list has 100000 elements, then we will have 100000 / 50 = 2000 groups, all at the same level.

The new version creates groups of 10 elements across each level, as you can see below.
image
This causes the expand buttons to react a lot faster to clicks.

### DIFF
--- a/utils/json.js
+++ b/utils/json.js
@@ -5,10 +5,24 @@ import {
 } from '@mdi/js'
 import _ from 'lodash'
 
-const groupSize = 50
-
 export default function itemifyJSON(jsonText) {
+  const groupsPerLevel = 10
   let id = 0
+  function minifyList(list, base = 0) {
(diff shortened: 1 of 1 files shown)


streitlua:
One thing I noticed is that the size of the groups are not consistent across levels, like in the following example:
image

I don't know if that's confusing or not, but the plus is that it allows to save clicks (we don't need to go down 5 levels if the groups contain a single element).

ffsinger (approved):
I think it's fine how you handle the tail of the list

fquellec (approved):
Yes ! Nicely done

Issue #265, opened 2021-11-29
Accessors for the file explorer


andreaskundig:
(I'm leaving this text as it is. Meanwhile bookmarks have been renamed to accessors, and newer comments have the latest information)

(This is what a remember from a discussion with @pdehaye in the train this morning. Unfortunately, it is now inextricably mixed up with my thoughts about how to name and implement it.)

Allow users to pinpoint interesting places in their data,
so they can find it again and share it.

A bookmark is made of

1.  the location of the file in the zip (a unix path)
2.  if needed: the location of the data in the file (jsonpath for json)
3.  the structure of the data (jsonschema for json?, maybe for csv too?)
4.  if needed: ui to display it, but maybe that can be inferred from its structure? (table, list, tree)

@emmanuel-hestia, maybe you have ideas for better formats to encode this data, I haven&rsquo;t looked into all the possibilities of json-ld etc&#x2026;

# Defining a bookmark

The file explorer ui lets the user set bookmarks. A node in the tree-viewer defines 1. and 2., we just need to add a button or a checkbox to add a bookmark.

We need and additional ui to define 3. and 4.

A list of json objects that all have some keys in common (ad impressions for one advertiser) could be displayed as a table. Hiding a column would indicate that attribute should not be shared. This data would have the type of an array of objects, which we could decide to display by default as a table. The exact jsonschema of the resulting json could be inferred with a library like [quicktype](https://github.com/quicktype/quicktype).

A single json object could be displayed with checkboxes next to each property.

It might make sense to give the user the option two switch between a tree and a table. Some lists will contain objects with children, or very disparate objects with few keys in common. Maybe there&rsquo;s a way to deal with that by having trees in table rows, but too much of that risks looking ugly. If the user can choose the ui, his choice must be indicated in part 4. of the bookmark. (Could this be some kind of annotation on the schema?)

(Note that if we define things this way, the place where data location ends and data structure starts is defined arbitrarily by the user. Hopefully this won&rsquo;t cause us too many headaches?)

# Bookmarks component / interaction with file explorer

The bookmarks themselves are displayed above the file explorer.
Clicking on a bookmark displays the bookmarked place in the explorer. Several bookmarks can be selected at the same time.

# Export / import

The selected bookmarks can be downloaded as a json, with or without the data they reference. The json can be included in the exported/encrypted zip file.

A bookmark json can be uploaded to an experience. If it contains only the bookmarks, display what&rsquo;s at the bookmarks in the current data.

[...]

PR #277, opened 2021-12-03
feat: individual files export

ffsinger:
Changes:
- Add checkboxes next to each file in the file explorer (but only if a consent form is defined for this experience)
- Add checkbox in consent form to include selected files in the export
- Import page recreates a file explorer with the included files
- Consent form shows the size of the zip before sending it

I'm using Vuex to store the names of the selected files, it's cleaner than using props since we are passing data to sibling components (UnitFileExplorer and UnitConsentForm).

Directly related issues:
- https://github.com/hestiaAI/digipower-data-experiences/issues/91
- https://github.com/hestiaAI/digipower-data-experiences/issues/88

Indirectly related issue:
- https://github.com/hestiaAI/hestialabs-experiences/issues/265

### DIFF
--- a/components/the-data-experience/TheDataExperienceDefault.vue
+++ b/components/the-data-experience/TheDataExperienceDefault.vue
@@ -49,12 +49,12 @@
       </VRow>
       <VRow v-if="showDataExplorer">
         <VCol>
-          <UnitFileExplorer v-bind="{ fileManager }" />
+          <UnitFileExplorer v-bind="{ fileManager, selectable: true }" />
         </VCol>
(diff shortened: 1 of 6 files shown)


streitlua (approved):
Looking good!
I tested the facebook, spotify, twitter and explorer experiences and everything seems to be working well!

As we discussed on signal, it would be great to help the users better understand what the checkboxes are for, because they can only understand that once they reach the consent form and read the option to share the selected files in the file manager.
We should discuss how to to that with the others, and I think that we can merge this for now.

PR #278, opened 2021-12-03
feat: added a summary of the files in the file viewer and moved it to the top

streitlua:
Changes:
- added a summary of the files uploaded by the user:
image
- moved file viewer to the top of the experience
- added an util function to convert a size in bytes into a human readable string
- added size of current file selection in file explorer if selectable

Related issues:
- (https://github.com/hestiaAI/digipower-data-experiences/issues/70)
- (https://github.com/hestiaAI/digipower-data-experiences/issues/71) 

Remarks:
- Alex mentioned showing the number of datapoints in the files, but this is impossible to define generically in my opinion.
What do you think?


valentinoli:
> * Alex mentioned showing the number of datapoints in the files, but this is impossible to define generically in my opinion.

I agree. It seems very subjective and difficult to compute and hard to maintain long term

components/unit/file-explorer/UnitFileExplorer.vue
```diff
@@ -1,70 +1,90 @@
...
-            <VCardText>
-              <VTreeview
-                v-model="selectedFiles"
-                dense
-                open-on-click
-                activatable
-                rounded
-                return-object
-                transition
-                :selectable="selectable"
```
ffsinger: The files are not selectable anymore, this is a merging mistake

components/unit/file-explorer/UnitFileExplorer.vue
```diff
@@ -145,6 +151,45 @@ export default {
...
+        v => v.length
+      )
+      const orderedCounts = _.sortBy(
+        Object.entries(occurrences),
+        ([_ext, count]) => -count
+      )
+      const showAtMost = 2
+      const topExtensions = _.take(orderedCounts, showAtMost)
+      const descriptions = topExtensions.map(
+        ([ext, count]) => `${count} ${ext}${count === 1 ? '' : 's'}`
```
ffsinger, line 180: I don't like this format: it's weird to see "3 jss" or "5 jsons", I'd rather write "3 .js" and "5 .json"

ffsinger (changes requested):
It's nice ! 

As we discussed, maybe we could also show the size of selected files

alexbfree:
Added some feedback; see hestiaAI/digipower-data-experiences/issues/71

valentinoli:
> As we discussed, maybe we could also show the size of selected files

You can borrow/change this function from another project of mine that converts bytes to a size string

```
bytesToSizeString (bytes) {
  if (!bytes) {
    // just in case
    return ''
  }
  const kilo = 1024
  const kiloBytes = Math.floor(bytes / kilo)
  let sizeString = `${kiloBytes} KB`
  if (kiloBytes >= kilo) {
    // express in MB if >= 1024 KB
    const megaBytes = Number.parseFloat(kiloBytes / kilo).toFixed(1)
    sizeString = `${megaBytes} MB`
  }
  // surround with parentheses
  return `(${sizeString})`
}
```

streitlua:
> You can borrow/change this function from another project of mine that converts bytes to a size string

Thanks! We have a similar function in `utils.js` already :)

fquellec (approved):
Everything looks good to me, some UI to fix later but that's not the priority.. Well done !

ffsinger (approved):

PR #282, opened 2021-12-06
feat: Netflix usage and notification graph

fquellec:
Added two visualizations to the Netflix experience: 
- The first one allows the user to get an overview of how they use Netflix, based on the day of the week, time of day, device used, user, etc. 
- The second is the number of notifications that Netflix sends to each user.

In this PR, I also fixed the following issue: https://github.com/hestiaAI/digipower-data-experiences/issues/101
- I added some date formats 
- I fixed a bug where CSVs with only one column were not displayed.

Note: If you don't have Netflix data to test, you can ask Alex to provide it.


fquellec:
Do not hesitate to give advice to the UI, if you want to try different color palettes, etc.

manifests/experiences/netflix/postprocessors.js
```diff
@@ -0,0 +1,57 @@
...
+
+    results.push(row)
+  })
+  // sort by date
+  results.sort((e1, e2) => e1.date - e2.date)
+  return { headers: ['date', users].flat(), items: results }
+}
+export default {
+  usagePerDay,
+  notifsPerDay
```
ffsinger, line 56: It would be better to have this code inside the custom pipeline instead of as a post-processor

Also, it seems that usagePerDay isn't used

fquellec: You are right, I have deleted usagePerDay. What should the postprocessor be used for? Here I am using a custom pipeline to get the data and a postprocessor to transform it into the correct format. (pivot the table and compute duration)

fquellec: okok understood ;)

ffsinger (approved):

PR #283, opened 2021-12-06
feat: snapshot testing and documentation

### LINKED ISSUE #185: Improve README
1. Start with general information on what the project is about, for people who are not familiar with it.
2. Then, explain the different deployements (experiences.hestialabs.org, test.hestialabs.org, ...)
3. Finally, give information for power users and developers (sections "Build setup", "Global configuration" and "Upgrading npm on Windows")

Let's remove the "Special directories" section, it was added by default by Nuxt. We can link just link to the [doc of nuxt](https://nuxtjs.org/docs/get-started/directory-structure) instead.

ffsinger:
Changes:
- Fixed a bug that caused some tests to fail (`RegeneratorRuntime` error)
- Added an example snapshot test
- Updated the README with some documentation about testing, the TFAC instance, and changed the overall structure.

Close #185 
Contributes to #191

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,18 +1,22 @@
 # HestiaLabs Experiences
 
+The experiences developed by HestiaLabs are targeted for any user willing to better understand their personal data. It takes the form of an easy-to-use web app that locally processes the raw personal data downloaded from various sources (Twitter, Facebook, Uber, etc) and creates meaningful visualizations.
+
 ## Instances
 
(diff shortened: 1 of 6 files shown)


ffsinger:
If no snapshot test is failing it means that the snapshots are up to date, so it's expected that nothing happens. It would just rewrite the same snapshots.

You only need to run this command when the test fail because you've changed the content of a component.

PR #287, opened 2021-12-07
feat: hourglass pointer and webworkers

andreaskundig:
- Shows the cursor as an hourglass when it is above the file explorer while a csv or  json file is loading.
- Uses workers to avoid blocking browser rendering, when parsing csv and json files and when building the twitter sql database.
- Removes the options to see all entries of a table on the same page.

[hestiaAI/digipower-data-experiences#67](https://github.com/hestiaAI/digipower-data-experiences/issues/67)

### DIFF
--- a/components/unit/file-explorer/UnitFileExplorer.vue
+++ b/components/unit/file-explorer/UnitFileExplorer.vue
@@ -8,6 +8,11 @@
     :min-height="height"
     height="auto"
   >
+    <style v-if="isFileLoading">
+      :root {
+        --cursor-style: wait !important;
+      }
+    </style>
     <VNavigationDrawer
       ref="drawer"
       :mini-variant.sync="mini"
@@ -102,10 +107,12 @@
             :is="componentForType"
             v-bind="{ fileManager, filename }"
             v-if="supportedTypes.has(fileType)"
+            @loading="onLoading"
           />
           <UnitFileExplorerViewerUnknown
             v-else
             v-bind="{ fileManager, filename }"
+            @loading="onLoading"
           />
         </template>
         <template v-else>
@@ -140,6 +147,7 @@ export default {
       mini: true,
       miniWidth: 48,
       search: '',
+      isFileLoading: false,
       height: 500,
       nDataPoints: null,
       sortedExtensionTexts: [],
@@ -246,6 +254,9 @@ export default {
   },
   methods: {
     plurify,
+    onLoading(loading) {
+      this.isFileLoading = loading
+    },
     setSelectedItem([item]) {
       // item might be undefined (when unselecting)
       if (item) {
@@ -332,6 +343,13 @@ export default {
   overflow: hidden;
 
   color: rgba(0, 0, 0, 0.87);
+}
+
+.explorer,
+.explorer .v-icon,
+.explorer .v-treeview-node__root,
+.explorer .v-treeview-node__root > .v-treeview-node__content > * {
+  cursor: var(--cursor-style);
 }
 
 .explorer__content {
--- a/components/unit/file-explorer/viewer/UnitFileExplorerViewerCsv.vue
+++ b/components/unit/file-explorer/viewer/UnitFileExplorerViewerCsv.vue
@@ -9,10 +9,11 @@
 
 <script>
 import mixin from './mixin'
+import mixinLoading from './mixin-loading'
 
 export default {
   name: 'UnitFileExplorerViewerCsv',
-  mixins: [mixin],
+  mixins: [mixin, mixinLoading],
   data() {
     return {
       csvText: '',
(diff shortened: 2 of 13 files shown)


components/unit/file-explorer/viewer/UnitFileExplorerViewerCsv.vue
```diff
@@ -28,11 +28,24 @@ export default {
         await this.getContentFromFilename(filename)
       },
       immediate: true
+    },
+    loading: {
+      handler(value) {
+        this.$emit('loading', value)
+      },
+      immediate: true
```
valentinoli, line 36: @andreaskundig 
I got the hourglass to work without the need for a timeout. The trick was to emit the loading event from a watcher with `immediate: true`

valentinoli, line 36: @andreaskundig 
I got the hourglass to work without the need for a timeout. The trick was to emit the loading event from a watcher with `immediate: true`

components/unit/file-explorer/viewer/UnitFileExplorerViewerCsv.vue
```diff
@@ -28,11 +28,24 @@ export default {
...
   methods: {
     async getContentFromFilename(filename) {
-      await this.setLoading(true)
+      this.loading = true
+
+      // small timeout to allow the navigation drawer transition to finish
+      // note: nextTick() will not work for that purpose
+      // because it only waits for the virtual DOM to update and not the page
+      // https://stackoverflow.com/a/47636157/8238129
+      await setTimeoutPromise(200)
```
valentinoli, line 47: This line can be removed and the hourglass will still appear. However, we need a timeout for the navigation drawer animation to finish.

components/unit/file-explorer/UnitFileExplorer.vue
```diff
@@ -148,9 +148,6 @@ export default {
       // @fileType should match the postfix of the Vue component name
       return this.selectedItem?.type
     },
-    cursorStyle() {
-      return this.isFileLoading ? 'wait !important' : ''
-    },
```
valentinoli: simplified a bit ... set a `v-if` on the `<style>` instead

components/unit/file-explorer/UnitFileExplorer.vue
```diff
@@ -94,14 +94,14 @@
...
-            @loading="onLoading"
             v-bind="{ fileManager, filename }"
             v-if="supportedTypes.has(fileType)"
+            @loading="onLoading"
           />
           <UnitFileExplorerViewerUnknown
             v-else
-            @loading="onLoading"
             v-bind="{ fileManager, filename }"
+            @loading="onLoading"
```
valentinoli: The VSCode linter does that

components/unit/file-explorer/viewer/UnitFileExplorerViewerCsv.vue
valentinoli: Note, I only implemented this for the CSV viewer. Please test this and see if it is ok. Should be moved to the mixin you created... and now I realize I went a full circle with this.

components/unit/file-explorer/viewer/mixin-loading.js
```diff
@@ -6,19 +6,29 @@ export default {
...
+      // https://www.hesselinkwebdesign.nl/2019/nexttick-vs-settimeout-in-vue/
+      // TODO avoid magic number for setTimeout
+      // On my machine 200 works and 100 doesn't,
+      // but what will happen on other machines?
+
+      // small timeout to allow the navigation drawer transition to finish
+      // note: nextTick() will not work for that purpose
+      // because it only waits for the virtual DOM to update and not the page
+      // https://stackoverflow.com/a/47636157/8238129
+      await setTimeoutPromise(300)
```
valentinoli, line 31: I realized I went a full circle with my fixes and so the mixin is useful mainly for the setLoading method which I renamed to startLoading since we can just do "this.loading = false" inside the components

The timeout is only needed to allow the navigation drawer transition to finish. Without it, the hourglass still works but the transition lags.

andreaskundig:
I removed the reviewers because discussing with Valentin we found a possibly better solution that I need to test

ffsinger (approved):
Great improvement, it's a good idea that we start using web workers for intensive computations

components/unit/file-explorer/UnitFileExplorer.vue
```diff
@@ -102,10 +107,12 @@
             :is="componentForType"
             v-bind="{ fileManager, filename }"
             v-if="supportedTypes.has(fileType)"
+            @loading="onLoading"
           />
           <UnitFileExplorerViewerUnknown
             v-else
             v-bind="{ fileManager, filename }"
+            @loading="onLoading"
```
valentinoli, line 115: I guess we don't need to listen to the loading event on `UnitFileExplorerViewerUnknown`

valentinoli:
Great work @andreaskundig 
Important improvement

PR #290, opened 2021-12-08
style: improved file explorer summary string and left panel

streitlua:
Addressing most of the comments in:
- (https://github.com/hestiaAI/digipower-data-experiences/issues/71)
- (https://github.com/hestiaAI/digipower-data-experiences/issues/74)

image

Changes:
- summary of files:
  - bold font to file type and number of occurrences
  - capitalized extensions
  - italicized file names
  - changed list format (more compact and simple)
  - `data points -> lines` for text files
  - `data points -> datapoints` elsewhere
  - summary doesn't disappear when a file is selected
- left panel:
  - search box always appears at the top
  - `100% -> 80%` width 
  - clicking on selected file unselects it

Some problems that remain (but that could be solved in a following PR):
- summary of files:
  - path of files could be shortened, but we cannot keep just the file name since we'll have duplicates (e.g. google my activity)
- left panel:
  - need to improve experience of going from it back to the main panel

### DIFF
--- a/components/unit/file-explorer/UnitFileExplorer.vue
+++ b/components/unit/file-explorer/UnitFileExplorer.vue
@@ -6,7 +6,7 @@
       :mini-variant-width="miniWidth"
       absolute
       permanent
-      width="100%"
+      width="80%"
     >
       <template #prepend>
         <VListItem class="px-2">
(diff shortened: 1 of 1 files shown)


ffsinger (approved):

alexbfree:
Saw the comment on duplicate file names.

How about we omit path name *except* where the top 3 files we want to
mention include duplicates, in which case do include the (possibly
abbreviated) folder name?

PR #316, opened 2021-12-10
feat: generic location viewer

fquellec:
New generic sub-experiment, takes a set of files and searches for data point coordinates in CSV and JSON files. Displays each data point in a Kepler map.  The viz can be improved but the main goal is achieved and you can try it in the "Other" experience, DEMO : 
image

In order to detect the coordinates, we have two different approaches: 
- for CSVs: we check that we have at least two columns with names containing 'lat' or ['lon', 'lng'], and for a sample of the values in these columns we check that all values are valid longitudes or latitudes. 
- for JSONs: we browse the object to find places where, at the same "level", we have two keys with names containing 'lat' or ['lon', 'lng'] and valid latitude and longitude values, once found, we return the current siblings as the description and the path to get there. 

Functions to validate latitude and longitude values : 
`const isValidLat = num => isFinite(num) && Math.abs(num) <= 90`
`const isValidLon = num => isFinite(num) && Math.abs(num) <= 180`


manifests/generic-pipelines.js
```diff
@@ -194,4 +194,125 @@ async function timedObservationViewer(fileManager, manifest) {
...
+    if (isObject(node)) {
+      // Object
+
+      // First check if there is possible lat and lon at this level
+      let latHeader = null
+      let lonHeader = null
+      Object.entries(node).forEach(([k, v]) => {
+        const kLw = k.toLowerCase()
+        // Specific to Google format, lat and lon are multiplied by 1e7
+        if (kLw.includes('e7')) node[k] = v = v * 1e-7
```
streitlua, line 213: I wonder if this mutation propagates to the `#jsonItems` dictionary of the `fileManager` and affects other experiences as well

components/chart/view/ChartViewGenericLocationViewer.vue
```diff
@@ -0,0 +1,71 @@
+<template>
+  <VContainer v-if="values.length > 0">
+    <VRow>
+      <VCol cols="12">
+        <p class="text-h6">Number of location records in your files</p>
+        <p v-if="total === 0" class="text-subtitle-2">
+          No location were found in your file(s).
```
streitlua, line 7: small type: "no location**s**"

manifests/generic-pipelines.js
andreaskundig: That's something to watch out for indeed, but the only call of this function is on line 304

`extractJsonLocations(JSON.parse(json))`

where json is obviously a string, so there's nothing to worry about

streitlua (commented):
Works well for me!
One thing that we could add is a file filter like we have in the generic date viewer?

PR #318, opened 2021-12-10
feat: select files in consent form

ffsinger:
Instead of having checkboxes in the file-explorer experience, we can open a dialog from the consent form. This dialog contains a file-explorer with checkboxes.

https://github.com/hestiaAI/digipower-data-experiences/issues/111

Screenshot_20211210_180658

Screenshot_20211210_180840

TODO:
- When we open a file that takes a lot of vertical space (e.g. an unwrapped json or an image) and open the navigation drawer, then the component does not shrink, it just shows blank space where the file was shown. (This was already the case before this PR)

### DIFF
--- a/components/TheAppBar.vue
+++ b/components/TheAppBar.vue
@@ -1,6 +1,6 @@
 <template>
   <div>
-    <VAppBar fixed app color="white" height="75" style="z-index: 2000">
+    <VAppBar fixed app color="white" height="75">
       <VAppBarNavIcon
         aria-label="Open navigation menu"
         @click.stop="drawer = !drawer"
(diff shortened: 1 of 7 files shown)


andreaskundig (commented):
I think it would be better to show only the file selector, and disable jumping to the file viewer. It feels a bit weird that when you want to choose what files you upload you can get to a button to download them.

It's not obvious what the CLOSE button does. It would be clearer if it was called OK. I would expect CLOSE to work like a CANCEL button that ignores what has been chosen.

ffsinger:
> I think it would be better to show only the file selector, and disable jumping to the file viewer. It feels a bit weird that when you want to choose what files you upload you can get to a button to download them.

True, but if the person does not remember what the file contains, it would be cumbersome to go back to the file-explorer experience and check what's inside.

> It's not obvious what the CLOSE button does. It would be clearer if it was called OK. I would expect CLOSE to work like a CANCEL button that ignores what has been chosen.

I changed it from SAVE to CLOSE because if you click out of the dialog it's also saved, and I found that counterintuitive. But OK works best so I'll change to that

PR #331, opened 2021-12-13
Feat/accessor

andreaskundig:
Unfortunately I couldn't write a test for the file-manager's findMatchingObjects()

The FileManager cannot be instantiated in node because the syntax for creating web workers is different from what works with webpack for the browser.


ffsinger (approved):
I don't think we really need the tests `jsonpath examples` and `ajv examples`, as they are only testing external functions, but for the rest it's good that we have tests

ffsinger:
> The FileManager cannot be instantiated in node because the syntax for creating web workers is different from what works with webpack for the browser.

Putting this at the top of the test file may solve the problem:

```
/**
 * @jest-environment jsdom
 */
```

PR #336, opened 2021-12-14
fix: removed computation of the number of data points temporarily

streitlua:
…since it causes crashes on the file explorer.

For me, on Google Chrome, this solved the crashes with Anaelle's facebook zip but not with Charles'.


ffsinger:
It still freezes on Chromium for me

fquellec:
Weird.. It is OS dependent then.. I'm on MAC OS

ffsinger:
I'm on Linux. Note that I don't have any issue with the same archive on Firefox

streitlua:
Also on Linux and getting exactly the same behavior as Florian

Issue #332, opened 2021-12-14
Bug with Google Experience


alexbfree:
Hi, I downloaded my Google data per the standard instructions we are giving to participants (I will attach contents list below)

However the file was not loadable in our Google experience. It caused the browser to hang for a minute or two 

![one](https://user-images.githubusercontent.com/1473244/145997490-42a34c96-b986-4dbe-b0b4-d0922b1b384d.png)

and even when I regained control, no data was displayed 

![two](https://user-images.githubusercontent.com/1473244/145997561-f73a0296-1e8f-4348-b212-015f23cbe125.png)

and eventually the page auto-reloaded, I guess because it couldn't complete rendering)

![three](https://user-images.githubusercontent.com/1473244/145997575-18673038-0d3f-407f-9d0e-5a757518d566.png)

ffsinger:
I think that the problem is not directly related to the Google Experience, but probably more to the fact the zip contains 5534 files and 1.65 GB

alexbfree:
That count is incorrect tho, the file was 175Mb

ffsinger:
There may be some compression going on then. 175MB compressed and 1.65GB when unzipped

alexbfree:
BTW here is an extract from the instructions we are using:

of these don't appear for you, it just means you don't use them so can ignore. You can also omit any that you don't wish to see the data for.
> 
>     Arts & Culture
>     Chrome
>     Crisis User Reports
>     Data Shared for Research
>     Fit
>     Google Account
>     Google Cloud Search
>     Google Help Communities
>     Google Pay
>     Google Play Books (options on this need changing - see step 6 below)
>     Google Play Games Services
>     Google Play Movies & TV (options on this need changing - see step 6 below)
>     Google Play Store
>     Google Shopping
>     Google Store
>     Google Translator Toolkit
>     Google Workplace Marketplace
>     Google Hangouts
>     Home App
>     Keep
>     Location History (options on this need changing - see step 6 below)
>     Maps
>     Maps (Your Places)
>     My Activity (options on this need changing - see step 6 below)
>     My Maps
>     News
>     Pinpoint
>     Profile
>     Purchases & Reservations
>     Question Hub
>     Reminders
>     Saved
>     Search Contributions
>     Tasks
>     YouTube (TWO options on this need changing - see step 6 & 7 below)
> 
> 5. You can add others that may appear for you that are not listed here if they are of interest, but take > care NOT to select (as they may be large or cause delays):
[...]

PR #337, opened 2021-12-14
fix: crash when unzipping on chrome

ffsinger:
Replaced unzipit with JSZip

https://github.com/hestiaAI/hestialabs-experiences/issues/332


ffsinger:
Ah the difference is that now we don't have the zip shown in the file explorer, it shows directly the folders that are inside.

It creates a small UI bug where the scrollbar is visible until we click on the sidebar for the first time:
Screenshot_20211214_180337

fquellec (approved):
No UI bugs for me, works well !

PR #339, opened 2021-12-15
refactor: change custom pipeline arguments

ffsinger:
Custom pipelines need various arguments, but most of the time not all of them. This PR allows a custom pipeline function to get the parameters it needs without caring about the order. It should be a bit less error-prone, especially if we add/remove some arguments in the future


streitlua (approved):
Code looks good and tested changes on FB, TC, Twitter, Uber and generic pipelines (not netflix)

Issue #286, opened 2021-12-07
Content Sharing with Web Share API


valentinoli:
Supersedes https://github.com/hestiaAI/data-pipelines/issues/27 (please read that issue first)

[Integrate with the OS sharing UI with the Web Share API](https://web.dev/web-share/)
[Receiving shared data with the Web Share Target API](https://web.dev/web-share-target/)

[Web Share API](https://developer.mozilla.org/en-US/docs/Web/API/Web_Share_API)
[`Navigator.share()`](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/share)

valentinoli:
Created a PR in the polyfill to support Facebook hashtag sharing 
https://github.com/on2-dev/share-api-polyfill/pull/79

For devices that actually support Web Share, I guess we need to simply append the hashtags to the shared text.

valentinoli:
Created a PR in the polyfill to support Twitter `via` parameter (which we used to provide to the deprecated `TwitterShareButton` component through the `twitter-user` prop)

https://github.com/on2-dev/share-api-polyfill/pull/80

valentinoli:
`dom-to-image-more` not working for an iframe (tried to export the Kepler iframe to a PNG)
https://github.com/1904labs/dom-to-image-more/issues/57

valentinoli:
> `dom-to-image-more` is not working for an iframe (tried to export the Kepler iframe to a PNG) and got an error
> [1904labs/dom-to-image-more#57](https://github.com/1904labs/dom-to-image-more/issues/57)

This is only an issue for us if we want the user to be able to share the Kepler map.

I'll close this issue for now and we can create new issues as problems arise or requirements change

andreaskundig:
> `dom-to-image-more` is not working for an iframe (tried to export the Kepler iframe to a PNG) and got an error [1904labs/dom-to-image-more#57](https://github.com/1904labs/dom-to-image-more/issues/57)

Did you try adding the library to stat/kepler/index.html  as a script ?

``` html
<script src="path/to/dom-to-image-more.min.js" />
```

andreaskundig:
There are now potentially several share buttons on a page:
- on the top right, for sharing a link to the page
- appears below every chart, after pressing export
- appears below every table, after pressing export

Since the webshare api is not implemented everywhere, the functionality differs according to browser and os. 


| OS  | Browser | experience |chart|table|
| ------------- |||||
[...]

PR #341, opened 2021-12-17
feat: Web Share (#286)

valentinoli:
Implemented a share button that uses the Web Share API with a [polyfill](https://github.com/on2-dev/share-api-polyfill/). 

[See issue](https://github.com/hestiaAI/hestialabs-experiences/issues/286)

I've implemented a `BaseShareButton` component. It can be used for file sharing too when the device/browser supports it. I added the share button to `UnitFilterableTable` to share the exported CSV of a data table, and to `UnitVegaViz` to share the viz as an image. The share button in the top right corner of an experience is just for sharing meta-info about the experience page - `url`, `title`, `text`. I append the `hashtags` to the `text`. These properties are also included when a file is shared but can be overwritten.

The Windows native share dialog is quite limiting. I know that the Twitter Windows app supports it (although CSV sharing does not seem to work). For Windows we might want to consider turning it off and [using the polyfill instead](https://github.com/on2-dev/share-api-polyfill/issues/69#issuecomment-863512457) but this we can discuss.

**Polyfill**
Note that the [polyfill](https://github.com/on2-dev/share-api-polyfill/) package does not support file sharing and has some bugs (I have submitted PRs for some) but it is pretty nice! 
My PRs:
https://github.com/on2-dev/share-api-polyfill/pulls/valentinoli


valentinoli:
might want to use this https://github.com/1904labs/dom-to-image-more

for the ChartView component to produce a single image

instead of extracting all the d3 vizualization SVG elements in the ChartView and creating an image for each one

ffsinger (changes requested):
You have updated all the dependencies in `package-lock.json` (e.g. webpack has been updated to version 5). 

I made the same mistake a few days ago in https://github.com/hestiaAI/hestialabs-experiences/pull/328, and fixed in https://github.com/hestiaAI/hestialabs-experiences/commit/e6adc803a1b9fec1a2dd5e0dd46487861842b94b

Otherwise, code looks good

valentinoli:
> You have updated all the dependencies in `package-lock.json` (e.g. webpack has been updated to version 5). 
> 
> I made the same mistake a few days ago in https://github.com/hestiaAI/hestialabs-experiences/pull/328, and fixed in https://github.com/hestiaAI/hestialabs-experiences/commit/e6adc803a1b9fec1a2dd5e0dd46487861842b94b
> 
> Otherwise, code looks good

Oh no! I had some issues with that... thanks for noticing.

ffsinger:
The button at the top right works fine for me, but the other "Share" buttons in the viz/tables are not visible on my machine (Linux, Firefox or Chromium), even though they are listed in the component hierarchy of vue-dev-utils. Do you have an idea of what could be the problem?

valentinoli:
You are right Florian that if there is no share button the export button is not needed but I found it clumsy and even more inconsistent to remove it... Also, the code becomes cleaner and the old workaround for the export+download in 1 click of the download button is not needed...

Downside is that the user needs to first export and then download...

ffsinger:
I agree that for the table it's fine to have the Export and Download as separate buttons if the share option is not possible. But for the visualizations it's confusing to have this export button that does nothing, e.g. for the twitter advertising overview it's like this:

Screenshot_20211220_112928

valentinoli:
Ok, that is not good

I have to take a better look

valentinoli:
@ffsinger 
Now the Export button should never appear on its own. I forgot to add a Download button beside it.
By the way, I created a component `ChartViewVRowWebShare` that is basically a `VRow` with the export/download/sharing functionality. Clicking "Export" is exporting the entire content of the `VRow` as an image.

ffsinger (commented):
`ChartViewSunburst` should also use `ChartViewVRowWebShare`

Otherwise it looks good now

ffsinger (approved):

valentinoli:
@andreaskundig your review is pending

andreaskundig:
@valentinoli I took this week off, but one approval is enough for a pull request, you can merge it.

PR #342, opened 2021-12-17
style: update file selection in consent form

ffsinger:
Changes: As described in #350, **except** for the following:

- The file content can still be browsed because it will soon be necessary for selecting individual data points. However, the file hierarchy is shown by default and clicking on a file doesn't automatically switch to this panel.
- I think that clearing the selection when the user unchecks the box would be a very frustrating experience (easy to misclick). Instead, there is a "Clear selection" button inside the dialog. The checkbox is automatically unchecked when clicking this button.

Note: the dialog has been moved to a separate component


fquellec (approved):
Very nice UI/UX, works well, just need to fix merging conflict !

PR #344, opened 2021-12-17
feat: table experience and custom audience view (facebook)

fquellec:
This PR is related to https://github.com/hestiaAI/hestialabs-experiences/issues/385 and https://github.com/hestiaAI/hestialabs-experiences/issues/349
Added a new type of pipeline based on accessors. We can create experiments by specifying : 
- a regex to match a set of files: `fileAccessorRegex`
- a jsonPath to access a set of elements in those files: `nodeAccessor`
- a jsonPath for each value/column in those elements: `field`

Here is an example we use for Facebook: 
```
{
      "key": "customAudience",
      "customPipeline": "jsonToTableConverter",
      "customPipelineOptions": { 
        "rootAccessor": {
          "fileAccessorRegex": "/ads_information/advertisers_using_your_activity_or_information.json",
          "nodeAccessor": "$.custom_audiences_all_types_v2[*]"
        },
        "properties": [
          {
            "name": "Advertiser",
            "field": "advertiser_name",
            "type": "string",
            "required": true
          },
          {
            "name": "has_data_file_custom_audience",
            "field": "has_data_file_custom_audience",
            "type": "boolean",
            "required": true
          },
          {
            "name": "has_remarketing_custom_audience",
            "field": "has_remarketing_custom_audience",
            "type": "string",
            "required": true
          },
          {
            "name": "has_in_person_store_visit",
            "field": "has_in_person_store_visit",
            "type": "string",
            "required": true
          }
        ]
      },
      "showTable": true,
      "title": "Advertisers using your activity or information",
      "text": "See all advertisers that use Facebook to get information about you"
    }
```

Note that the customPipelineOptions are validated with a JsonSchema, and we could do this for the entire manifests.

### DIFF
--- a/manifests/experiences/facebook/facebook.json
+++ b/manifests/experiences/facebook/facebook.json
@@ -66,7 +66,7 @@
       "customPipeline": "jsonToTableConverter",
       "customPipelineOptions": { 
         "rootAccessor": {
-          "fileAccessorRegex": "/ads_information/advertisers_using_your_activity_or_information.json",
+          "fileAccessorRegex": "ads_information/advertisers_using_your_activity_or_information.json",
           "nodeAccessor": "$.custom_audiences_all_types_v2[*]"
         },
         "properties": [
@@ -85,13 +85,13 @@
           {
             "name": "has_remarketing_custom_audience",
             "field": "has_remarketing_custom_audience",
-            "type": "string",
+            "type": "boolean",
             "required": true
           },
           {
             "name": "has_in_person_store_visit",
             "field": "has_in_person_store_visit",
-            "type": "string",
+            "type": "boolean",
             "required": true
           }
         ]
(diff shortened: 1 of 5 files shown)


pdehaye:
To insist, because easily missed: `field` could be a JSON path, not just an attribute. So we are really decomposing the accessing into several bits (to a file, to a "semantic" node, to an attribute downstream to that semantic node, the latter not necessarily just one down)

fquellec:
> To insist, because easily missed: `field` could be a JSON path, not just an attribute. So we are really decomposing the accessing into several bits (to a file, to a "semantic" node, to an attribute downstream to that semantic node, the latter not necessarily just one down)

Yes, it's already a JsonPath, the above example just doesn't need to go any further down in the hierarchy.

ffsinger (approved):

Issue #403, opened 2021-12-21
fix: Bug in Twitter experience pipeline


fquellec:
Targeting criteria are not associated to the right advertiser

PR #404, opened 2021-12-21
fix: #403

fquellec:
Fixed two issues with the Twitter SQL pipeline: 
- Tweet ids should be interpreted as TEXT, otherwise it causes overflows.
- The database ids were not unique, thanks to @ffsinger for his help :)


ffsinger (approved):
Thanks !

Issue #366, opened 2021-11-25
Ability to Share a specific datapoint (node+any children) and its sourcefile, path, data etc


alexbfree:
(excuse the brevity, adding tickets in a rush. Contact Alex or Paul for more details)

button always present on every node

to allow us to learn and collect in context examples.

also with ability to add comment (to build up a knowledgebase)

ffsinger:
It seems that the UX I was planning to implement (putting the node selection as part of the file selection dialog, for a finer-grained sharing) is not what you have in mind . So should it be part of the file explorer experience, or as a separate dialog (where?) ?

Issue #350, opened 2021-12-15
Fix some UX issues in "Select file" popup


alexbfree:
The file selection dialog here should be much simpler and more task-focused than the standard File Explorer view. It ONLY exists to select files for sharing, and has no other capability.

1) Hide the Summary. 
2) Remove the Details Panel - hierarchy should always be on screen
3) If top level has just one node, open the first level of children to save the user having to do it.
4) Make sure nothing except select/unselect happens when you click a filename (currently it tries to preview file)
5) it should be impossible to collapse/hide the tree of files (in which case, no need for a save icon, and the tree's container can be centred not left justified.
6) the text entry "Selected Files (n selected)" needs to be *within* the tickbox list, with its own tickbox (too easy to miss this choice otherwise). 
7) It should be impossible to leave the Select Files option selected with 0 files selected, and impossible to leave it unselected if more than 0 files are selected. Which implies:
 - when you tick the checkbox, that triggers the dialog, 
 - if you cancel the dialog, the checkbox becomes unchecked
 - if you select file first without having checked the checkbox, and select at least one file, the checkbox becomes ticked when you exit the dialog.
 - unchecking the dialog when files are selected has the effect of clearing the current selection. 

Small tweak just next door so I will combine into this ticket:

8) The "drop file here" link should be moved to a large button, same size as the Download File link. Also please add a (1) and (2) (maybe each as white text on a coloured circle?) within the button text of both buttons to indicate the two steps.

alexbfree:
> > Remove the Details Panel - hierarchy should always be on screen
> 
> For issue https://github.com/hestiaAI/digipower-data-experiences/issues/78 we need this panel

This should be two different variants of the dialog. Selecting Files is not the same cognitive task, nor is it done at the same time, as selecting a specific node for sharing. Selecting specific nodes for sharing should be done in the context of when you are reviewing a file... selecting files is done when you are using the purposes form.
Probably we need an additional entry in the purposes form for "selected datapoints" which shows a count/popup of whatever was shared above. I realise this implies sharing some data between the file explorer and the purposes form so we need some careful design here. 
In terms of the specific issue raised here, the slight awkwardness when selecting files, I will review current UX once deployed and comment more.

PR #419, opened 2021-12-22
Improvements and bug fix for file selection

ffsinger:
Addresses the following comments (via signal):

- why is the popup so wide? I have massive amounts of whitespace at right. Could the width be dynamically decided?
- it feels odd that the file explorer is a separate frame 'within' the select files dialog. any way to make it a bit more integrated?
- File explorer title not needed here.
- As discussed in #350 the details panel and < arrow should be hidden. This dialog should ONLY be for selecting files. Let's not incorporate any of #366 solution into this, that would probably be a separate dialog.
- Editing your selection is a bit weird. I think it would be better if "2 selected" was a link to trigger the dialog (without changing the checkbox).

The last point lets us work around a bug where the "Select files" button did not appear in the instances (but did appear in a local environment).

Based on https://github.com/hestiaAI/hestialabs-experiences/issues/350#issuecomment-999515601, I have decided to stop using the whole FileExplorer component in this popup, because we only need a very small fraction of it.

### DIFF
--- a/components/TheDataExperience.vue
+++ b/components/TheDataExperience.vue
@@ -40,7 +40,7 @@
         <template v-if="success">
           <VRow v-if="showDataExplorer">
             <VCol>
-              <UnitFileExplorer v-bind="{ fileManager, selectable: false }" />
+              <UnitFileExplorer v-bind="{ fileManager }" />
             </VCol>
           </VRow>
           <VRow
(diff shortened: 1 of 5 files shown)


andreaskundig:
This looks much better than the last time I've looked at it, good job! 

The search field in the file selection jumps up and down while you type because the size of the dialog depends on the number of the files found. That's slightly unexpected, but let's leave it like this until someone complains.

Issue #348, opened 2021-12-15
Bug: It can be painful for participants to locate / be guided to the right View within an Experience


alexbfree:
We experienced a problem when we were guiding the participants through using the experience (them in control and screensharing, us guiding them over Zoom) - we found ourself saying "up a bit, down a bit, all the way to the bottom, just above the table..." etc. etc. 

One solution i propose is simply that we give the Views simple named tabs left to right, with only one tab on screen at a time.

- Tab 1 could be Summary (from file explorer)
- Tab 2 could be Files (the File Explorer)
- Tab 3 could be File Viewer (whichever file is currently behing viewed)
- Tab ? would be Data Over Time (dated observations viewer)
- Tab ? would be Map View (generic geolocation viewer)
- Tabs ? would be company specific views 
- Final tab would be "Share my data" (purposes/consent form)

Tab approach should be responsive - use the screen width on desktops+(tablets in landscape) but maybe handled differently on phone+(tablets in portrait) to make use of vertical space.

@pdehaye please add any comments/alternative suggestions

valentinoli:
Pull Request has been submitted for review

PR #424, opened 2021-12-25
fix: use VTabs for UI of experience (#348)

valentinoli:
Resolves https://github.com/hestiaAI/hestialabs-experiences/issues/348


components/TheDataExperience.vue
```diff
@@ -1,53 +1,68 @@
 <template>
-  <client-only placeholder="Loading...">
-    <keep-alive>
```
valentinoli, line 3: client-only and keep-alive can be removed, they are old remnants

ffsinger (approved):

PR #434, opened 2022-01-03
fix: and refactor data validation

### LINKED ISSUE #406: Empty SQL results considered as invalid by viz and tables
If we have a SQL query that returns nothing, the result will be an object with empty headers and empty items. The data tables and viz consider this as invalid, because they expect the headers to be non-empty. The expected behavior would be to show `No data found` instead of `Data in this format cannot be displayed`.

There are two possible solutions:

1. (Preferred) In the function `select` of `sql.js`, find a way to return the headers even if the result is empty (e.g. by parsing the query ourselves).
2. In the table/viz: relax the `isDataValid` function

ffsinger:
Refactored the data validation into a single component `DataValidator`, because the same conditions were checked in three different places (`ChartView`, `UnitVegaViz` and `UnitFilterableTable`).

There are three cases:
- The format is invalid: a warning is shown
- There is no data: an info box is shown
- Otherwise: show the viz or table

Also slightly relaxed the condition for having valid data: headers can be empty. In this case, it will be considered as having no data. This closes #406


fquellec (approved):
Nice ! Lot more clean like that, small question, wouldn't be even easier/possible to call DataValidator only once UnitQuery.vue ?

ffsinger:
> wouldn't be even easier/possible to call DataValidator only once UnitQuery.vue ?

It's a possibility, but we have these problems:
- The tables and graph components can be used outside of UnitQuery (e.g. the file explorer or export page), in which case we would have to put the data validator in all these parents
- UnitIframe does not use this validation, and if we have to handle this special case in UnitQuery it's a bit complex

Issue #335, opened 2021-12-14
[Bug] Twitter experience: top advertisers graph shows incorrect number of advertisers


streitlua:
If you uncheck the tick "display others", the number of shown advertisers is incorrect: (here 2 instead of 3)
![image](https://user-images.githubusercontent.com/16099301/146019988-7a56deda-324a-4da7-916c-27e3b4ef62d8.png)

I imagine that this occurs because the filter for others is done after the selection of the top n advertisers, so to solve this we would just need to invert the order of the computation.

Issue #405, opened 2021-12-21
fix: Twitter top advertiser graph


foucault-dumas:
![signal-2021-12-21-155326_001](https://user-images.githubusercontent.com/81250365/146959839-90d6cde4-6a2c-41a2-a418-5996c3fb7977.png)

- [ ] add here a missing information in this experience: "You have been targeted by XXX ads between 15/04/2021 and 13/06/2021" --> this very important (and impressive) number never appears.
- [ ] remove "in total" in this sentence, it's confusing
- [ ] solve the "per day" display, misleading and not understandable as it is

PR #436, opened 2022-01-04
feat: top row graph + scoped style for all graphs

fquellec:
- Added a new chart for top-k bar charts to replace the old vega one. You can test it on the Twitter experience (top advertiser) : 
image
- Fixed several issues we had with overlapping styles between graphics using <style scoped>, this introduced a lot of problems with tooltips but they should all be fixed with this PR. Note that for style inside SVG content we need to prefix them with ```::v-deep``` otherwise they are not applied when used with <style scoped>.

### DIFF
--- a/components/chart/ChartView.vue
+++ b/components/chart/ChartView.vue
@@ -61,3 +61,6 @@ export default {
   }
 }
 </script>
+<style>
+@import 'assets/styles/dc.css';
+</style>
--- a/components/chart/view/ChartViewGenericDateViewer.vue
+++ b/components/chart/view/ChartViewGenericDateViewer.vue
@@ -309,11 +309,11 @@ export default {
   }
 }
 </script>
-<style>
-.range-chart > svg > g > g.axis.y {
+<style scoped>
+::v-deep .range-chart > svg > g > g.axis.y {
   display: none;
 }
-.hide {
+::v-deep .hide {
   display: none;
 }
 </style>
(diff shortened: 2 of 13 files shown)


ffsinger (approved):
Cool !

It's not very clear what the "average" aggregation does. Maybe change the x axis to "Count / day" in this case, or add a small explanation.

Can you reuse this graph for the facebook experience ? It's still using vega

fquellec:
- I forgot to mention that this issue is related to #405.
- For the "average" aggregation, it doesn't make much sense anyway, I wanted to remove it but I need @foucault-dumas to confirm that it's ok to remove it? 
- For the Facebook experiment, I'll change the graph! Thanks for your review @ffsinger

ffsinger:
It's also related to #335 (I didn't check if it was fixed though)

> For the "average" aggregation, it doesn't make much sense anyway, I wanted to remove it

I agree

fquellec:
Yes it fixes #335 too, need to double check though

streitlua (changes requested):
The graph looks good for me and indeed seems to fix #335, but the graph is a bit too large for my screen (it doesn't fit vertically), is there any way to make it smaller or adaptive?

fquellec:
I pushed your recommendations, could you merge and close the PR @streitlua if you agree with the changes?

streitlua (approved):
That's better! I would decrease the font size of the x axis a bit but that's just my personal taste. I will merge it like this.

foucault-dumas:
Thanks guys for all this.
I don't manage to see these changes on test.hestialabs.org nor on experiences.hestialabs.org 🤔

Issue #411, opened 2021-12-22
Investigate feasibility of adding XLS/XLSX support


alexbfree:
We would not need a full embedded Excel viewer. Is there some library that would allow us to read in XLS/XLSX files, interpret them as CSV, and render through our existing CSV to table logic?

If it's easy please do it as part of this ticket. If it's a lot of work please add comment about how much work and we can assign priority accordingly.

andreaskundig:
One library that can parse excel files: https://github.com/SheetJS/sheetjs

alexbfree:
sounds like it would be worth someone doing a proof of concept to see how much work integrating that library would be. I see it can pick tables from HTML too which might also be useful.

valentinoli:
https://www.npmjs.com/package/read-excel-file

alexbfree:
Seems like it is tabulating OK, but not picking up column titles

<img width="1177" alt="Screenshot 2022-01-27 at 17 10 13" src="https://user-images.githubusercontent.com/1473244/151408703-86d5847d-ae07-4735-9dc3-99286c9e6938.png">

PR #437, opened 2022-01-04
feat: XLSX viewer

ffsinger:
#411 

Add a viewer for XLSX files in file explorer. 

I'm using https://www.npmjs.com/package/read-excel-file, but it doesn't handle well huge excel files ([see this comment](https://github.com/catamphetamine/read-excel-file/issues/38#issuecomment-544286628)). If this is a problem, we could use https://github.com/SheetJS/sheetjs instead.


fquellec (approved):
Works like a charm for me :)

streitlua (approved):
Works well for me as well!

Issue #407, opened 2021-12-22
FileExplorer: some elements of the viewer stay over the navigation drawer


ffsinger:
In the file explorer, when we open the navigation drawer, it happens that some elements of the right panel stay visible.

Example: open a CSV, then open the navigation drawer. The headers of the table appear among the files.

![Screenshot_20211222_095713](https://user-images.githubusercontent.com/25420002/147065666-83820d04-12d2-4725-8b51-381f3aa952d6.png)

alexbfree:
Good that you have spotted this bug. @pdehaye and I noticed this in a deep dive but didn't manage to capture a screenshot.

PR #439, opened 2022-01-05
fix: z-index issues

ffsinger:
#407

The table header no longer goes over the navigation drawer. I did it by setting the z-index attribute.

At the same time, I'm suggesting to put the summary string in an expandable panel, as illustrated below, because it can take a lot of screen space. It is open by default, and the user can close it when they start exploring the files content.

Screenshot_20220105_103529

Screenshot_20220105_103559

We can revert this commit if you don't like it

### DIFF
--- a/components/TheAppBar.vue
+++ b/components/TheAppBar.vue
@@ -1,6 +1,6 @@
 <template>
   <div>
-    <VAppBar fixed app color="white" height="75">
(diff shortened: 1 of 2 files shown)


streitlua (approved):
It's working well for me, good job!
And I definitely think that being able to hide the summary is a good improvement ;)

PR #444, opened 2022-01-07
refactor: results passed to consent form

ffsinger:
Put the results in the store instead of passing them to a sibling component via events and props (`UnitQuery` -> `TheDataExperience` -> `UnitConsentForm`).

Vuex detected that some d3 graphs modify the result objects (e.g. `CharViewSearchSunburst`), so I cloned them before passing them to the graphs.

Also removed some old forgotten code.

I still find the consent form code to be quite messy (especially `UnitConsentFormSection`) but I'm not sure how it could be improved.

### DIFF
--- a/components/TheDataExperience.vue
+++ b/components/TheDataExperience.vue
@@ -77,14 +77,12 @@
                   index,
                   vega
                 }"
-                @update="onQueryUpdate"
               />
             </VTabItem>
             <VTabItem v-if="consentForm">
               <UnitConsentForm
                 v-bind="{
                   consentForm,
-                  allResults,
                   defaultView,
                   fileManager,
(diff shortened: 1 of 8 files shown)


streitlua (approved):
Seems to be working correctly.
One thing that I noticed (that is maybe not related to this PR) is that the experience on the import page does not use tabs.

ffsinger:
> One thing that I noticed (that is maybe not related to this PR) is that the experience on the import page does not use tabs.

It's not related to this PR, but you're right. If we want to have the tabs in the import page, we could simply copy the new code for tabs, or find a better way to factorize this common code between `import.vue` and `TheDataExperience.vue`

Issue #365, opened 2021-11-25
Make the Process and Run buttons hidden and automatic


alexbfree:
(excuse the brevity, adding tickets in a rush. Contact Alex or Paul for more details)

Low priority, but we think these buttons add extra clicks and are not needed. All data viewers should populate once data is loaded.

ffsinger:
Now that the experiences are separated in tabs (#424), we have multiple options:

- Run all the experiences automatically when the data is loaded. This can cause a long initial delay, and maybe the user is only interested in one specific sub-experience
- Run an experience automatically when the user switches to the tab. This reduces the problem above. However, what should happen if the user clicks on another tab before the current one has finished loading ? options: 
  - Kill the computation
  - Queue the computations
  - Do not let the user change tabs

alexbfree:
Yes, I like the second option, run the experience when you switch to the tab.

I would reject queuing the computations, I don't think there are benefits for doing that.
This leaves the other two options of killing the computation or blocking tab change.

I don't have strong views on this, I guess killing the computation would result in a more responsive UI so would be slightly preferably if it is easy to do without side effects. If that is difficult or carries other impacts then blocking the tab change would be OK.

valentinoli:
I think killing computation would be much harder to implement - unless the computation is done in a worker, then there is no way to interrupt the thread. But if the computation is done in a worker, I think it can be done simply by [`worker.terminate()`](https://developer.mozilla.org/en-US/docs/Web/API/Worker/terminate)

valentinoli:
@ffsinger 
Did we consider using web workers for sql.js?

valentinoli:
The best temporary solution in my view is to just disable the tabs and show a progress indicator. I implemented an overlay with a progress indicator that should only appear if a subexperience takes long.

ffsinger:
> Did we consider using web workers for sql.js?

Yes it's planned to use web workers for the sql queries at some point

andreaskundig:
I reverted the merge of #455 because it broke the implementation of #400 that allows the user to edit pipeline options. After changing the options, the user needs to be able to run the pipeline again, but #455 removed the run button.

valentinoli:
@andreaskundig you mean #445 ?

andreaskundig:
Yes, I corrected in the comment

valentinoli:
@andreaskundig so what is the plan? We still want this feature, right?

andreaskundig:
[...]

Issue #400, opened 2021-12-20
Ability to edit, test modification of a view and send to developers via a button


alexbfree:
scenario:
Paul or Alex or Charles wants a new view of some specific data.
1) Using existing file explorer experience they can select a specific file and click a button to get its accessor
2) Using existing file explorer experience they can select a specific file and click a button get its fieldname (within that path)
3) Using some kind of advanced view, they can create a new view (probably starting with an existing view as template), view the source configuration of that view, edit it/tweak it a bit (perhaps using grabbed data from 1 & 2?) and then switch back to rendered view to test it. 
4) when happy, they can click a "send to developers" button which emails the configuration file (with a comment box text) to development (perhaps some shared alias we set up, or store on Kdrive or somewhere via API + send notification?)

@pdehaye @andreaskundig @fquellec This is my first attempt at specifying the minimal 'view editing functionality' we discussed in standup today. I am certain this first draft can be improved - please feel free to make edits or add comments.

pdehaye:
This is too complex already _for a first step_. 

The accessor can be selected using [https://jsonpath.com/](https://jsonpath.com/) , the formatting of the schema can be obtained using [https://app.quicktype.io/](https://app.quicktype.io/). This is enough to enable many more of us to create new, relevant, tables. Can @emmanuel-hestia be the first to test that workflow, to specialize that generic view for one particular company? He has, after all, all the Shex maps at the ready.

fquellec:
As a first step, I suggest we build a sub experience/card where we can try/edit the configuration files and test them on the uploaded content. So we would have a json editor for the configuration and a run button, when the user fills in the configuration file and clicks run, a filterable table is displayed or error messages if any. Now, where should we place this type of map, in an advanced view of each experiment or in a standalone page, like the playground?

alexbfree:
Updated based on update given in standup today

PR #445, opened 2022-01-07
fix: auto-run subexperiences

valentinoli:
Addresses #365

### DIFF
--- a/components/TheDataExperience.vue
+++ b/components/TheDataExperience.vue
@@ -25,13 +25,13 @@
           <span>Processing files...</span>
         </template>
         <template v-else-if="error || success">
-          <VAlert
+          <BaseAlert
             :type="error ? 'error' : 'success'"
             border="top"
             colored-border
             max-width="600"
             >{{ message }}
-          </VAlert>
+          </BaseAlert>
         </template>
       </VCol>
     </VRow>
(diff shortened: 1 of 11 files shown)


valentinoli:
I'd like to suggest after this PR has been reviewed and merged that we move `UnitSql.vue` into the `unit` folder and rename pipeline components like so:

`UnitSql.vue` -> `UnitPipelineSql.vue`
`UnitSparql.vue` -> `UnitPipelineSparql.vue`
`UnitCustomPipeline.vue` -> `UnitPipelineCustom.vue`

and then move them to folder `unit/pipeline/`

ffsinger (approved):
Nice, good job !

I noticed that the overlay does not show up for SQL pipelines, but it might be because we are not using web workers for them yet ?

valentinoli:
> I noticed that the overlay does not show up for SQL pipelines, but it might be because we are not using web workers for them yet ?

It has nothing to do with web workers. The overlay only shows if the computation takes long time. If it is short, then no overlay shows - otherwise the overlay would flash in an annoying way.

Currently, I think this is our best solution. Terminating web workers can be a later solution if we want.

ffsinger:
> The overlay only shows if the computation takes long time

But if you try the twitter targeting criteria experience with the big dataset, it does take a long time and doesn't show the overlay.

valentinoli:
> But if you try the twitter targeting criteria experience with the big dataset, it does take a long time and doesn't show the overlay.

Yes, I noticed and I don't know why. I think it happens in Chrome and not on Firefox for example. Generally, I have noticed that the UI lags more in Chrome than it does in Firefox for some reason. So I think what you described might be related to that. Otherwise, it might be something else I overlooked

andreaskundig:
This breaks #400. When you change the options of a pipeline, you can no longer execute them :(

andreaskundig:
I reverted the merge

Issue #375, opened 2021-12-15
Across all data tables, ensure Timestamps are converted to human-readable dates


alexbfree:
This was observed with ad impressions data on Facebook experience. Let me know if you need sample data.

alexbfree:
Also seen with Samsung data.

alexbfree:
Just noticed this issue is present in JSON trees too - not sure if that needs handling separately. Example:

![Screenshot 2022-01-17 at 08 56 13](https://user-images.githubusercontent.com/1473244/149738634-f4e3bfef-c824-40dd-a207-d3ccd0d5ba50.png)

alexbfree:
@andreaskundig can you confirm if extra dev work is needed to fix in the JSON case - If yes, please take a look (or see if other devs can). If no, this can be closed again.

andreaskundig:
Yes it does need extra work

PR #467, opened 2022-01-11
fix: #375 format field to dates when header contains "date" or "time" substrings

fquellec:


ffsinger (changes requested):
I tried with the facebook file `your_off-facebook_activity.json` (experience "Ad interactions") and it doesn't work as expected. The timestamp in the file is given as a standard Unix timestamp (number of seconds passed from 1970). But a Javascript Date expects a timestamp representing the number of _milliseconds_ passed from 1970 ([see here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date#the_ecmascript_epoch_and_timestamps)). So you should multiply `value` by 1000 **if** it looks like a unix timestamp

fquellec:
Yes, as we discussed this morning, this is a quick workaround needed for tomorrow's deep-dive. Many date formats are possible and we could detect them as we do with the generic date viewer, but I think doing this every time we post a table could slow the process down unnecessarily. For example how do you detect timestamps in milliseconds or seconds (both are possible) and without detecting phone numbers ?

ffsinger:
I'd say that the timestamp in milliseconds is much less likely than in seconds, because the former seems to be only used in javascript.

If it's really needed for tomorrow we can merge it like this, but it's really confusing to have all these dates show up as 1970 in the Facebook experience

fquellec:
um in my experience I seem to encounter more timestamps in milliseconds than in seconds but you're right it's weird to have all these dates displayed as 1970, let's discuss this during tomorrow's meeting!

andreaskundig (approved):

ffsinger (approved):

Issue #409, opened 2021-12-22
Create a test for an sql pipeline


andreaskundig:
The feedback loop for creating an sql pipeline is too slow because we have to run it in the browser. It would be much less tedious to be able to run it on the command line.

Create a first test for any pipeline. Once we have one that works, it will be easy to create new ones for other pipelines when needed.

Issue #233, opened 2021-11-23
SQL DB and RDF Store are exported as a global instance


ffsinger:
As discussed in the dev meeting, we may want to have multiple instances of these objects in the future. So they shouldn't be exported directly as a new instance accessible from anywhere in the code.

valentinoli:
Maybe a factory could be exported and the instances for the user stored in Vuex?

andreaskundig:
I was also thinking that we're not using vuex enough

valentinoli:
@andreaskundig I assume this one can be closed

PR #468, opened 2022-01-11
test: add tests for twitter sql database

ffsinger:
#409

Some refactoring was needed for these tests:
- Because Jest does not like web workers, I had to remove the worker that we were using in `databaseBuilder` for twitter. I first tried mocking the workers to avoid doing this, but it doesn't work because the the test suite crashes during the import, before we can even mock them. Actually, it does not affect performance because the slow part of `databaseBuilder` is when we _insert_ the rows in the db, and this part was not run in the worker anyway.
- Removed the global export of the DB (#233). Now a new DB is created when we run `databaseBuilder`.

Added the following tests:
- Check that `databaseBuilder` inserts correctly all the rows
- Check that the query `advertisers-per-day` returns the correct items

For these tests we are using a very simple mock of `ad-impressions` and `ad-engagements`, so that we have only a couple of rows.

I still need to write tests for the other twitter queries, but I'd like to have a review of this first.


andreaskundig (approved):
excellent!

Issue #460, opened 2022-01-10
Hide unnecessary path info in summary


alexbfree:
Path information is making summary strings somewhat unreadable, for example:

![Screenshot 2022-01-10 at 15 48 18](https://user-images.githubusercontent.com/1473244/148795873-8988ef32-871b-4ed5-8bae-3752860d3ac1.png)

Please simplify the view per these rules:
1. show ONLY filenames UNLESS the filename is non-unique (disregarding path/depth) across the whole uploaded files list 
2. If Non-unique filename include ONLY the immediate parent folder of that file (and potentially add grandparent/greatgrandparent - only add enough levels as you need to make a UNIQUE string)

PR #473, opened 2022-01-12
feat: hide unnecessary path info in summary

ffsinger:
As described in #460


andreaskundig (approved):
There's no test for setShortFilenames, but I trust you did it by hand at least.

ffsinger:
Yes I tested manually, but you're right, I should have added one. I'll do it later

Issue #408, opened 2021-12-22
Ability to click a button next to a JSON node 'View as Table'


alexbfree:
As proposed by @andreaskundig in standup today and in #346, the idea is to have a button next to a JSON node that says "view as table". This follows on from #346 and #391 and #400 is an achievable alternative to supporting tables-within-treeview. 
It will open up a table view underneath the treeview (or on top in a popup? see what looks best)
Important that the focus moves to the new table when you click.

andreaskundig:
As a first step I added jsonPath to every node that's displayed in the json tree view.

Next we'll have to generate a glob path for the file. The file name needs to be escaped in case it contains characters with a special meaning in the glob syntax. See the [micromatch library](https://github.com/micromatch/micromatch) we're using.

alexbfree:
Just confirming this ticket remains super valuable, would have loved to have it in deep dive today!

ffsinger:
It's already deployed though. When you click on the text of a node, the table appears below the file explorer

alexbfree:
Can someone provide instructions - i didn't find this feature

andreaskundig:
You can click on any json node in the the viewer, and a table will appear under the explorer

alexbfree:
I will play with it

fquellec:
Done in https://github.com/hestiaAI/hestialabs-experiences/pull/474
Needs Improvement but part of a bigger issue

PR #474, opened 2022-01-13
Feat/json table

andreaskundig:
This essentially implements #408 and is ready for feedback. 

However, in the spirit of #400, the table that appears should have an editable configuration, so that the display can be improved the accessors modified, and the whole configuration cut and pasted into a manifest to define a permanent table.

(Note that the new code uses previously unused code that I already committed on master.)


ffsinger (commented):
It breaks the JSON search: now only the leaf node are hidden if they don't correspond to the filter. I don't know which change is causing this

fquellec (approved):
good to me ! I think there is room for UX improvement but let's discuss this with Alex once it's merged and deployed on test.

andreaskundig:
> It breaks the JSON search: now only the leaf node are hidden if they don't correspond to the filter. I don't know which change is causing this

Glad you caught that, I'll have a look at it. Are things still working correctly on the master branch?

ffsinger:
Yes it works fine on master

andreaskundig:
I merged this to master so we can deploy it to get some feedback, as discussed

PR #475, opened 2022-01-13
chore: deploy experiences.hestialabs.org

ffsinger:
- Add all the latest changes
- Add tracker-control


andreaskundig (changes requested):
I tried merging into netlify-test with a pull request the other day. It created a commit on netlify-test just for the merge, which made the two branches diverge. Maybe there's a way to avoid that problem when merging in github, but I ended up doing it from the command line. 

Otherwise, yes please update the netlify branch

ffsinger:
Ah yes I see, I'll do it from the command line then

ffsinger:
For next time I think that we can just select "Rebase" when merging the pull request (https://stackoverflow.com/a/43940548)

andreaskundig:
It's not encouraging that they called it rebase, that sounds like rewriting history.

We could try that and see and check that hash of the commits remains the same on both branches.

ffsinger:
You're right, they rewrite the commit SHAs (https://stackoverflow.com/questions/60597400/how-to-do-a-fast-forward-merge-on-github). Let's not do this

Issue #461, opened 2022-01-10
Suppress common Mac, Windows temporary files


alexbfree:
Please ignore files and folders such as `__MACOSX` (folder) and `.DS_Store` - and whatever equivalents occur on Windows (+Linux? +Android?) upon import

PR #478, opened 2022-01-13
fix: ignore MacOS tmp files

ffsinger:
#461 

Tell me if you know other files that should be ignored


fquellec (approved):
That's all I think

streitlua:
In the Uber data, there is a file `.DS_Store` so maybe we should do a regex match that is case unsensitive?

ffsinger:
Good point, it's fixed

streitlua (approved):
Looks good to me!

PR #489, opened 2022-01-17
fix: xlsx crash

ffsinger:
The latest refactoring on `UnitFilterableTable` broke the XLSX viewer

Related to #411

### DIFF
--- a/utils/json.js
+++ b/utils/json.js
@@ -110,6 +110,10 @@ export default function itemifyJSON(jsonText, filter) {
 }
 
 export function formatObject(object) {
+  if (!object) {
+    // null is of type object
+    return object
+  }
   const keys = Object.keys(object)
   if (keys.length === 0) {
     return '{no attributes}'


andreaskundig (commented):
After having merged this because it's good that it fixes the xslx viewer, I wonder:  shouldn't formatObject return a string? If the xlsx viewer displays just converts this to a string "null", we should be doing it here explicitly, for example with  `''+object `

ffsinger:
Hmm if we want the UnitFilterableTable to show a blank space on null values, we need to pass either null or an empty string. If it becomes the string "null" it will show up in the table and we don't want that.

So if we really want the format function to return a string, it should be the empty string. Should I make this change ?

andreaskundig:
Ok for the empty string

ffsinger:
Fixed in 4c0b1a75c5073e6756319de64d79574ad1cf3d6e

PR #493, opened 2022-01-17
fix: undefined values in twitter sql

ffsinger:
Some users have data with ads that don't have an `impressionTime`. This was causing an error when inserting the values into the database, because we didn't properly replace `undefined` values with `null`

I've fixed the bug and added a new test that checks for these cases.


fquellec (approved):
Nice !

Issue #480, opened 2022-01-14
Show which files are expected in each experience - and show what's wrong when mismatched.


alexbfree:
Standalone feature that will allow other tickets such as #353 to be done later.

Each experience should communicate to user which files it is expecting WITH EXACT PATH info.
This info needn't be on screen all the time (either collapsible side panel or popup from an (i) icon)
It is most needed at the dragging file/just about to press Process Files stage.

The task is to (a) develop the feature to show such info on screen
(b) to go through all experiences and map whatever internal specification/files to check for list - and expose that into the feature - so that the experience user can now precisely what paths/files are being looked for.

Also: When a file is not found. please show what WAS found.

e.g. 
```
expected:
data/ad_impressions.js
but found
twitter-username/data/ad_impressions.js
```
The idea here is not to do anything clever like in #353 - just show what we expect, and when not found, list alternative locations of those files that are present, with the goal of helping the user to fix their zip structure, (even if the need to do that is ultimately superseded by #353).

While #353 may take time/be blocked, this ticket should move ahead faster to give us a better UX right away while we wait for a longer term fix.

alexbfree:
As discussed in standup today, even tho #353 has been addressed, we still want this info available as it is good UX for users.

We realised in standup discussion that two things are needed
- at upload time, information to help you decide what is worth uploading
- on specific views, information about which files are needed/used/being included in that view

alexbfree:
Confirmed working nicely on digipower

alexbfree:
wait, that might have been premature. will reopen for now

alexbfree:
In Google experience, I see this, which is not right:

<img width="600" alt="Screenshot 2022-01-27 at 16 52 52" src="https://user-images.githubusercontent.com/1473244/151405893-56ca7491-a9cd-4430-9765-3a43667ccbb4.png">

alexbfree:
Also in Google experience, I have one dataset where I am getting this. I am not getting any information on why the data is not being found. I have checked and it is definitely JSON data files present - must be some other reason. 

<img width="974" alt="Screenshot 2022-01-27 at 16 49 27" src="https://user-images.githubusercontent.com/1473244/151405985-0ef3593b-4d9d-4d6b-b2dc-24be0c782be7.png">

[...]

Issue #353, opened 2021-12-07
Change the approach for loading files/restricting files/finding needed files


alexbfree:
As a general approach for all experiences (I speak for digipower, Charles/Paul can confirm whether this applies globally, I expect it does) we should not be restricting what files people upload. At the moment some experiences (such as Netflix) are locked to Zip and some others (such as Spotify) are locked to 1 file.

The general idea is people drop whatever they have (maybe the original zip, maybe some subset of files from what they unzipped, maybe just a single file) and we try and make sense of it.

If we can't find a file that we need (as an initial assumption we can go by the filename or path, but we may need to be cleverer than that in future) then we should just give a meaningful error message.. we couldn't display this view because we didn't find this file that we need.

In fact, we may want to consider presenting a list of "expected files/files we will use" for each experience (perhaps hidden under an info (i) popup or something) at the top of every experience.

alexbfree:
Discussed in standup today. 
Agreed that @alexbfree will make a ticket for just the last part. The rest is deferred pending other developments on refactoring/accessors [Sorry, I can't remember if it was @fquellec or @andreaskundig who was due to make a ticket for the outstanding work this depends upon]. 

Once those two new tickets exist I will put this one on hold.

alexbfree:
Incidentally we have now had 2 participants share their twitter data in this format of `twitter-YYYY-{longID}/data/` subdirs so I guess the format they deliver in has changed? (or maybe different for SAR vs download portal). Anyway, this shouldn't matter once this is fixed.

ffsinger:
#498 relaxed the restrictions for **ZIPs**. Now the files needed for Twitter, Facebook and Uber can be at any level in the archive.

One point that was not addressed is that we are still expecting a zip, so the user cannot just upload one unzipped file for these experiences. I suggest that you open a new issue if this is still needed, so that this ticket can follow the standard process (merged -> deployed to test -> deployed to digipower)

alexbfree:
OK I have created #501.

@ffsinger @andreaskundig @fquellec please can you all confirm it is OK to close this ticket once #498 is in production

(I previously got the impression there may be some other outstanding refactoring work, hinted at previously, that requires this ticket to stay open)

alexbfree:
On digipower currently - Uber is not finding the `trips_data.csv` if it is not in a folder called `Rider` in an `Uber Data` top level folder. Not sure if we expect this should have been fixed yet.

alexbfree:
[...]

PR #498, opened 2022-01-17
feat: relaxed file restrictions

### LINKED ISSUE #479: Facebook preprocessing
As discussed in the thread of #374, Facebook files are badly encoded and should to be passed through a preprocessor. However, since we don't know all the filenames that Facebook can give us, we need to allow regex / glob for mapping filenames to preprocessors. Two approaches are possible:

1. Have only one preprocessor function per company, applied to all files, and match the relevant filenames inside the function. This has been implemented in #476
2. Keep the mapping in the manifests, but add support for regex / glob

The team was in favor of approach 2, so we decided to keep this as it is because it is a low priority issue.

ffsinger:
Partly address #353 

Now the files that are required in a zip archive can be defined as globs. 

The Twitter and Facebook pipelines are adapted so that the files that we process can be at any depth in the archive.

Note that the preprocessing also had to be adapted for more flexibility, now it can also use globs. Close #479

I'm not sure if we should still show to the user the list of files that are required (#480). It may give the impression that we are only using these files, while in practice we are using all those that we can process (in the generic date/location viewer). In my opinion, it would be more informative to give this information in each experience (i.e. show "this experience uses the files ad-impressions.js and ad-engagements.js", somewhere not too visible).

### DIFF
--- a/components/TheDataExperience.vue
+++ b/components/TheDataExperience.vue
@@ -248,11 +248,19 @@ export default {
       const start = new Date()
 
       await this.fileManager.init(uppyFiles, this.multiple)
-      let processedFiles
-      try {
-        processedFiles = await this.fileManager.preprocessFiles(this.files)
-      } catch (error) {
-        this.handleError(error)
+
+      // Check that the required files are present in the archive
+      const missing = this.files
+        .map(filePathGlob => [
+          filePathGlob,
+          this.fileManager.findMatchingFilePaths(filePathGlob).length
+        ])
+        .filter(([_, n]) => n === 0)
+        .map(([filePathGlob, _]) => filePathGlob)
+      if (missing.length > 0) {
+        this.message = `Missing file(s): ${missing.join(', ')}`
+        this.error = true
+        this.progress = false
         return
       }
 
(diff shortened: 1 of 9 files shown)


fquellec (approved):
I agree with you about the list of files that are needed, As discussed with @alexbfree I think the solution is somewhere in between, we could have a file info button on each sub experience to inform on what files are used for that specific graph/table AND a file info button that summarizes the files that are needed for all experiments next to the upload module. In the latter, you could simply add at the end a message like "For ** and **, the tool will search all the files you have uploaded". Does that sound good ? Those messages will induce a lot of refactoring in the manifests, tell me if you need help with that :)

streitlua (approved):
I agree with what @fquellec said and with what was said in the meeting this morning. 
Tested the Twitter and Facebook experiences and all seems to work correctly. 
It would indeed be good to understand why some sub-experiences show "No data found" (for instance Anaelle's data for Facebook's "Advertisers using your data" sub-experience).

ffsinger:
Perfect, I agree with what you said. I'm going to merge this PR and continue working on this

Issue #481, opened 2022-01-14
feat: table view of json node is configurable


andreaskundig:
Table view of json node from #408 should be configurable like #400, so that the display can be improved the accessors modified, and the whole configuration cut and pasted into a manifest to define a permanent table.

The configuration involves an accessor to specify what to display and properties to specify what columns to display in what format. The syntax of the properties should probably be simplified.

PR #499, opened 2022-01-17
Feat: simplify table options

andreaskundig:
To make working with configurable tables as in #481, I propose simplifying the syntax to configure tables.

Maybe you can find a way to simplify them even further.

### DIFF
--- a/manifests/experiences/facebook/facebook.json
+++ b/manifests/experiences/facebook/facebook.json
@@ -72,7 +72,7 @@
           "filePath": "ads_information/advertisers_who_uploaded_a_contact_list_with_your_information.json",
           "jsonPath": "$.custom_audiences_all_types_v2[*]"
         },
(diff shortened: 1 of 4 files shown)


ffsinger (approved):

streitlua:
If I remember well, @fquellec and @pdehaye have come up with this specific format after some discussion, so it would be good to have at least @fquellec's input on the changes before we merge.

fquellec:
Yes, I don't mind simplifying the options, the `required` field for example is not useful at the moment, as for the `type` field on the other hand, it could/is already super useful in our current implementation, let's say you have an unrecognised date in your file, how/where do you need to specify the format for it to be parsed correctly? Another example, if you have very large numbers or floats to represent, we could let the user specify the desired format with the `type` and `format` fields. We could, however, make these fields optional.

PR #500, opened 2022-01-18
feat: add filters for columns in tables

fquellec:
We can now filter each column individually in addition to the global search, I have removed the old version where we could select the columns affected by the global search. It also fixes the date search which did not work as expected before.
Demo: 
image
image


ffsinger (commented):
It works, but it becomes weird with narrow columns:

Screenshot_20220118_111235

and in this case you can't see the filters either:

Screenshot_20220118_111559

streitlua:
Good point @ffsinger.
A possible workaround would be to show only a "search" button on top of each column, and then this button would open a popup of a much larger size with the text input and the checkboxes

fquellec:
Good point! I've already tried with the search button on each header but it's getting weird to have a popup within a popup... I'll set a minimum width and see how it goes!

streitlua:
I think I explained it badly, I didn't mean there to be a popup inside a popup, I just meant moving the search input text to the popup, and increasing the width of the popup.

fquellec:
No, you've explained it well. It's just that the search input text includes a popup for suggestions, so when you put it in another popup, the size of the main popup has to change when you start searching for something. But I can look into it!

valentinoli:
Did you consider just having a single input where you can select on which column you want to filter?

alexbfree:
One question - in the dropdown, are the checkbox-items ordered?
Intuitive order to me would be from most-commonly-occurring to least-common, in descending order? (Sorry could probably answer this myself but haven't had time to check yet.)

fquellec:
What do you think about this ? 
image

fquellec:
> Did you consider just having a single input where you can select on which column you want to filter? It would make the UI less cluttered.

Yes I agree but we need the ability to filter several column at the same time !

ffsinger (approved):
Yes it's much better, good job !

Just update the snapshot one more time before merging

PR #504, opened 2022-01-18
fix: twitter database where not computed properly

fquellec:
Looking at the `ad-engagements.js` and `ad-impressions.js` files, we concluded that the total number of ads that were shown to a person is the sum of these two files since there is no unique identifier linking the ads in these files. However, on closer inspection of the contents of these files, every ad with a tweet_id that is included in `ad-engagements.js` appears to be also included in `ad-impressions.js` but with a slightly different `impressionTime` (a few seconds). We conclude that these ads are the same but with the wrong or miscalled `impressionTime` attribute and this PR corrects the SQL database accordingly. Now, to detect engagement, we check for ads in `ad-impressions.js` that are similar to other ads in `ad-engagements.js` by checking that : 
- `advertiserName` is the same
- `displayLocation` is the same
- `tweetId` is the same
- `impressionTime` is close enough (120 seconds maximum difference)

We also added the `displayLocation` attribute to the database for future improvements ( classify ads by displayLocation )


ffsinger (approved):
Thanks for the fix !

fquellec:
Thank you for the test and refactor ;)

PR #505, opened 2022-01-19
fix: random crashes

ffsinger:
These are tentative fixes for crashes that I couldn't reproduce reliably. I'm not sure if every change is really needed, but it seems to fix (or at least reduce) the crashes. I don't have time to look at it more carefully because it should be merged quickly for the deep dives.

The crashes happened mostly in the File Explorer, when you had already opened files in another company (e.g. Tinder) and you go to another one (e.g. Tracker Control). It seems that the file manager and the store were left in a dirty state

### DIFF
--- a/components/TheDataExperience.vue
+++ b/components/TheDataExperience.vue
@@ -198,11 +198,7 @@ export default {
       success: false,
       message: '',
       rml: '',
-      fileManager: new FileManager(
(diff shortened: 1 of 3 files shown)


fquellec (approved):
All good I couldn't reproduce the issue

Issue #501, opened 2022-01-18
Remove restrictions on upload types/counts from all digipower experiences


alexbfree:
Following #353, #498, #345, #399, #480 we should remove all blocks on what people can upload. We should not insist on 1 zip file or N files etc. 

The preferred approach is upload what you have, we try to find what we need in it (which we communicate and tell you when not found (see #480)). People can upload zipped or unzipped, all or subset. it's up to them. We just display/interpret what we can from what's there.

PR #506, opened 2022-01-19
feat: show which files are required and relax restrictions

ffsinger:
The goal of this PR is to address #480. It adds icons near the file file upload component, and near each sub-experience. When clicking on the icon, a dialog opens and shows which files are expected/required.

To this end, I had to change how we refer to the files in the pipelines. Now, we identify the files of interest with an ID in the manifest. Example for Twitter:

```json
"files": {
    "impressions": "**/ad-impressions.js", 
    "engagements": "**/ad-engagements.js"
  }
```

The IDs are `impressions` and `engagements`, and they each map to a glob. Thanks to this, we will only refer to this ID in the different places where we use the files, instead of copy-pasting the glob. Note that we pass this object to the file manager.

The `UnitFiles` component uses this list to tell the user which files are expected:

Screenshot_20220119_152622

When clicking the icon:

Screenshot_20220119_152644

This icon also shows up in sub-experiences:

Screenshot_20220119_152514

Screenshot_20220119_152555

To know which files are required by a sub-experience, we need to indicate it in the manifest (using the IDs):

```json
"defaultView": [
    {
      "key": "advertisers-per-day",
      "sql": "advertisers-per-day",
      "files": ["impressions", "engagements"],
...
```

For working with these files in the pipelines, multiple methods have been added to the file manager:

```js
getFilePathsFromId(id, unique = true)
async getPreprocessedTextFromId(id, unique = true)
async getCsvItemsFromId(id, unique = true)
```

In most cases, we are expecting only one file. But if we want the list of all files matching the glob, we can use `unique=false`.

**Important note:**

The restrictions for uploading files have been further relaxed. Now there is no requirements for the upload (except the file type). The responsibility of checking that the correct files are present is passed to the sub-experience. For example, if we don't provide the file `ad-impressions` for twitter, most experiences will show a warning saying that this file was not found. The advantage is that the user will still be able to use the other experiences (file explorer, generic time viewer, etc).


ffsinger:
I couldn't test Netflix because I don't have the data, can someone check that it works ?

fquellec (changes requested):
Very Nice job, It works well ! Just one suggestion, I would replace the sentence : "All JSON/CSV files provided are analysed for dates and geographical coordinates. The related experiences will be more interesting the more files you include" By something like, "For the experiences marked with [insert file logo in red or other color], all JSON/CSV files provided will be analysed for dates and geographical coordinates. The related experiences will be more interesting the more files you include." And in addition to that add a [file logo in red or other color] to each generic experience repeating the information that all files are analysed.

ffsinger:
I just made a fix for the Facebook sub-experience that we define through an accessor.

It raises a very interesting question. A natural extension to the functionality presented in this PR would be to replace the raw globs with accessors. For instance, for twitter, we could have

```json
"files": {
    "impressions": {
        "filePath": "**/ad-impressions.js", 
        "jsonPath": "$.*.ad.adsUserData.adImpressions.impressions[*]"
    }, 
    "engagements": {
        "filePath": "**/ad-engagements.js", 
        "jsonPath": "$.*.ad.adsUserData.adEngagements.engagements[*].impressionAttributes"
    },
}
```

@andreaskundig you will probably want to comment on this

fquellec:
> I just made a fix for the Facebook sub-experience that we define through an accessor.
> 
> It raises a very interesting question. A natural extension to the functionality presented in this PR would be to replace the raw globs with accessors. For instance, for twitter, we could have
> 
> ```json
> "files": {
>     "impressions": {
>         "filePath": "**/ad-impressions.js", 
>         "jsonPath": "$.*.ad.adsUserData.adImpressions.impressions[*]"
>     }, 
>     "engagements": {
>         "filePath": "**/ad-engagements.js", 
>         "jsonPath": "$.*.ad.adsUserData.adEngagements.engagements[*].impressionAttributes"
>     },
> }
> ```
> 
> @andreaskundig you will probably want to comment on this

Thanks for the fix, I totally agree with this!

fquellec:
This PR also fixes https://github.com/hestiaAI/hestialabs-experiences/issues/501

ffsinger:
> Very Nice job, It works well ! Just one suggestion, I would replace the sentence : "All JSON/CSV files provided are analysed for dates and geographical coordinates. The related experiences will be more interesting the more files you include" By something like, "For the experiences marked with [insert file logo in red or other color], all JSON/CSV files provided will be analysed for dates and geographical coordinates. The related experiences will be more interesting the more files you include." And in addition to that add a [file logo in red or other color] to each generic experience repeating the information that all files are analysed.

Thanks for the suggestion, it's been done in the latest commit. I kept the icon black because red would have been a bit aggressive in my opinion

ffsinger:
> This PR also fixes #501

Not quite. We are still forcing the user to upload a certain file type in some experiences (a ZIP, JSON, CSV, etc)

fquellec (approved):

PR #507, opened 2022-01-19
fix: convert timestamps to date strings in json viewers + improve date detection

fquellec:
This PR is related to #375 

Add date translation feature to json viewer + improve code quality + improve date detection in both tables and json viewers.

To detect dates we try to convert the value with the Javascript Date constructor, we check if this result gives a valid date and then to make sure it is not a timestamp in seconds we check that the year is not 1970 (starting year of UNIX timestamps). If all checks pass, we transform the date into a local date string format, otherwise we return the value unchanged. This solution will fail only for dates from 1970, but this does not represent many use cases for us.


ffsinger (approved):
All good !

Issue #511, opened 2022-01-20
Unify the UI of the application


fquellec:
The application is made up of a large number of elements that were not meant to be displayed as they are now. We should take the time to unify each component of the user interface to something meaningful overall. For example, each sub-experience sheet was displayed in a list before and now in tabs, it seems a bit odd to have tabs with different heights.

I'm not going to list all the UI/UX issues here but rather send you some before/after screenshots to be discussed in this issue

fquellec:
First proposal: 
- make the whole experience page as tabs
- remove the shadow effect from the sub-experience cards
- display the newsletter as a disposable alert (and in the footer)
- make the share button at a fixed position. 
We still need to find a way to display the title of the experience (Netflix, Twitter, ...).

Demo: 
![image](https://user-images.githubusercontent.com/17878016/150355507-1116a6c1-ca49-416e-910e-e63aaef9f8fe.png)

alexbfree:
This looks great. Happy for this to be worked on as priority 2. :-) Sounds like it is nearly done.

alexbfree:
I agree with @ffsinger that it should not be called upload. I suggest "Load your Data"

alexbfree:
> We still need to find a way to display the title of the experience (Netflix, Twitter, ...).

If you haven't already solved this, I propose putting it where the "Analyze your private data" title is.
That title could change to e.g. "Analyse your Netflix data" with Logo just to left of title.

fquellec:
How about putting the title and logo of the experience in the top bar? That way we would always know which experiment we are in!
![image](https://user-images.githubusercontent.com/17878016/150509884-37bc878c-9a01-4564-a799-cc174864a5a6.png)

PR #517, opened 2022-01-21
fix: UI refactoring

fquellec:
This PR is related to #511 and apply the following changes: 
- make the whole experience page as tabs
- remove the shadow effect from the sub-experience cards
- display the newsletter as a disposable alert (and in the footer)
- make the share button at a fixed position. 
- Add title experiences in the app top bar

Demo: 
image

A lot can be done to improve the UI/UX, this is just a first step.
Now that the size of the sub-experience containers has changed, we need to improve the rendering of each sub-experience, but this could be done in separate issues.

### DIFF
--- a/components/TheAppBar.vue
+++ b/components/TheAppBar.vue
@@ -16,6 +16,18 @@
           <VIcon>$vuetify.icons.mdiHome</VIcon>
         </VBtn>
         <VSpacer />
+        <div class="d-flex">
+          <VImg
+            max-width="30"
+            :src="company.icon"
+            :lazy-src="company.icon"
+            contain
+          />
+          <h3 class="ml-3">
+            {{ company.title }}
+          </h3>
+        </div>
+        <VSpacer />
         <a
           v-if="collaborator"
           :href="collaborator.url"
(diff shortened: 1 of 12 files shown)


ffsinger:
The first part of the message on the alert should also be adaptive, because it's only The Eyeballs that is about the attention economy

Screenshot_20220121_154521

And personally I think that having the alert show up every time you refresh the page is a bit annoying, maybe we could keep the state in a cookie / localstorage, and show it back only once per day or something

fquellec:
> and show it back only once per day or something

I agree !

ffsinger (approved):

valentinoli:
Only problem with adding more things to the app bar is it doesn't take into account smaller screen sizes, on which I do not know how much emphasis we are putting... How important is it that the app looks good on mobile? For most apps, the importance is high.

image

alexbfree:
> And personally I think that having the alert show up every time you refresh the page is a bit annoying, maybe we could keep the state in a cookie / localstorage, and show it back only once per day or something

I agree, we probably don't even want it at all for Digipower, certainly not once a day.

ffsinger:
This could be configured by adding a parameter in https://github.com/hestiaAI/hestialabs-experiences/blob/master/config/digipower.json, so that we don't show it on digipower

Issue #514, opened 2022-01-20
Improve Twitter experience to allow filtering by follower-lookalike / advertisers and move between the two


alexbfree:
In the same way we can move between Tracked Apps and Targeted Advertisers on Tracker Control experience, we need to be able to do something similar on Twitter.

example scenario:

1. select a specific follower lookalike (maybe from lookalikes tag cloud or 1-advertiser-pie-chart)?
2. jump to the "Twitter advertising overview" but we are now filtering on only this follower lookalike (across all data). Alternatively there is a way to directly filter on a lookalike within Twitter advertising overview
3. can now pick on a specific advertiser that is targeting based on this follower lookalike from list and filter on that. Now we see only data for this lookalike-advertiser pair
4. now we can remove the original follower lookalike but still locked onto the advertiser and that is now the filter in effect - can see all lookalikes they used.

Other scenario would be 2-to-4 the other way round, starting with advertiser, then picking lookalike, then removing advertiser.

Possibly best way to do this is another graph onto the Twitter advertising overview (this would address 2-to-4 and 4-to-2).
I realise the jump described in step 1 are new - but would be useful if possible. That part is nice to have. Priority is to be able to do 2-to-4 and 4-to-2 via Twitter Advertising Overview.

We could also consider a new separate view for this. But that may be overkill.

alexbfree:
@pdehaye Hopefully this captures what we said. Please comment if you want any changes to the spec.

alexbfree:
@foucault-dumas you may wish to comment as well as this will be useful for TFAC

foucault-dumas:
The narrative could be important and is not taken into consideration in our experience.
Potential partners are interested in who is targeted by a particular narrative --> see the promoted tweets and take them as an entry point for a sum of data sets —> this promoted tweet was shown to w, y, z = X persons (x%) in our sample of people, using this and this criteria.

alexbfree:
I see this has been Production Deployed. Could someone just summarise what is now available and where?

fquellec:
I have added a search function in the Twitter overview for targeting criteria, advertiser and targeting value. This allows every interaction described in this issue.

PR #518, opened 2022-01-21
feat: add text filtering for targeting criteria and company in Twitter overview

fquellec:
This PR is related to #514.
To improve the graph's ability to explore different combinations of targeting criteria against advertisers, I have added a search function in the Twitter overview for targeting criteria, advertiser and targeting value. 

image

This way, a user can, for example, search for any targeting value that are not already displayed on the top 10 and see which advertiser uses that targeting criteria and vice versa. See the example below: 
image

In addition, I have added to the table the ability to attach a formatter for each column, this is used in TwitterOverview to convert engagement values 0 and 1 to 'Yes' or 'No'.

### DIFF
--- a/components/unit/filterable-table/UnitFilterableTable.vue
+++ b/components/unit/filterable-table/UnitFilterableTable.vue
@@ -47,7 +47,7 @@
                 dense
                 class="pa-4"
                 label="Search ..."
-                :items="columnValues(header.value)"
+                :items="columnValues(header)"
                 :menu-props="{ closeOnClick: true }"
               >
                 <template #selection="{ item, index }">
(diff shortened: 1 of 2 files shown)


components/unit/filterable-table/UnitFilterableTable.vue
```diff
@@ -76,11 +76,21 @@
...
+          )"
           #[`item.${header.value}`]="slotProps"
         >
           {{ formatItemAsString(slotProps) }}
         </template>
+        <template
+          v-for="header in headers.filter(header =>
+            header.hasOwnProperty('formatter')
+          )"
+          #[`item.${header.value}`]
```
ffsinger, line 90: This line should be ```#[`item.${header.value}`]="{ value }"```

Otherwise everything is formatted as "Yes" :)

ffsinger (changes requested):
The filters are not formatted:

Screenshot_20220121_175752

components/unit/filterable-table/UnitFilterableTable.vue
fquellec: Thanks ;) It's fixed now  !

ffsinger (approved):

PR #523, opened 2022-01-25
fix: remove all restrictions on file upload

ffsinger:
Changes:
- Add the functionality defined in #501. This can create some ambiguities in the pipelines about which files we need to use (i.e. what if the user puts two different twitter archives?), so here's how they handle it:
  - Tracker control takes all CSV found, and concatenates them if they have the same structure. Otherwise it takes the first one found.
  - Tinder takes the first JSON found (according to [the tinder shex](https://github.com/hestiaAI/data-catalog/blob/main/shex/tinder.shex) there is only one JSON in the archive)
  - Twitter/Facebook/Uber take the first file found
- Remove obsolete options from manifests (`ext`, `multiple`, `isGenericViewer`, `showDataExplorer`)
- Remove obsolete options from file-manager and refactor, with a slight functionality change (`removeTopmostFilenames` has been replaced by `removeZipName`, see docstring)
- Add found files in sub-experience dialogs

Note for the future: To better handle the ambiguities, we have the following options:
- Try to merge the files
- Show a warning to the user saying which file we are using
- Ask the user to choose the file they want to use

### DIFF
--- a/manifests/experiences/facebook/database.js
+++ b/manifests/experiences/facebook/database.js
@@ -14,7 +14,9 @@ export default async function databaseBuilder(fileManager) {
     ['timestamp', 'INTEGER']
   ])
   const advertisersInteractedWithFile = JSON.parse(
-    await fileManager.getPreprocessedTextFromId('advertisers-interacted')
+    (
+      await fileManager.getPreprocessedTextFromId('advertisers-interacted')
+    )[0] ?? null
   )
   const advertisersInteracted =
     JSONPath({
@@ -35,7 +37,9 @@ export default async function databaseBuilder(fileManager) {
   /// Advertisers who purchased your contact ////////////////////////////////////////////////////////////////////////////////
   db.create('advertisersContactInformation', [['name', 'TEXT']])
   const advertisersContactInformationFile = JSON.parse(
-    await fileManager.getPreprocessedTextFromId('advertisers-contact-list')
+    (
+      await fileManager.getPreprocessedTextFromId('advertisers-contact-list')
+    )[0] ?? null
   )
   const advertisersContact =
     JSONPath({
@@ -60,7 +64,8 @@ export default async function databaseBuilder(fileManager) {
   ])
 
   const offFacebookActivityFile = JSON.parse(
-    await fileManager.getPreprocessedTextFromId('off-facebook-activity')
+    (await fileManager.getPreprocessedTextFromId('off-facebook-activity'))[0] ??
+      null
   )
 
   const offFacebookActivityJSON =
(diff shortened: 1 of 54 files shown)


ffsinger:
(After our discussion yesterday I changed my mind and think that squashing before merging to master is a good idea, so we may want to start doing this. What do you think?)

utils/file-manager.js
```diff
@@ -288,46 +275,39 @@ export default class FileManager {
...
   async getPreprocessedTextFromId(id, unique = true) {
     const paths = this.getFilePathsFromId(id, unique)
-    if (paths === null) {
-      return null
-    }
     if (unique) {
+      if (paths === null) {
+        return null
+      }
       return await this.getPreprocessedText(paths)
```
andreaskundig, line 311: It's a bit confusing that the variable "paths" refers to just one path.

It works because getFilePathsFromId(id, true) returns a string, while getFilePathsFromId(id, false) returns an array.

I would like it better if these methods always returned the same type, an array. When unique=true, it would just return an array of length 1 at most.

andreaskundig (commented):
It's all good and works well, nice job!

However I would like it better if the file manager methods getXXXFromIds also returned an array when unique=true. That would make it easier to understand what they do, and would simplify the code where they're called.

utils/file-manager.js
ffsinger: > That would make it easier to understand what they do

I agree

> would simplify the code where they're called.

This I don't agree, for instance in the facebook pipeline it becomes:
`(await fileManager.getPreprocessedTextFromId('advertisers-interacted')[0]) ?? null`
instead of
`await fileManager.getPreprocessedTextFromId('advertisers-interacted')`
for the same result (if there is no match we need null instead of undefined, otherwise JSONPath will crash)

> When unique=true, it would just return an array of length 1 at most.

If we always return an array, then `unique` is not really useful because the pipelines can ignore the rest of the array without additional code. We can remove it in this case

So in summary it's kind of a trade-off between the simplicity of the API and the conciseness in the usage. I'm happy to do as you suggest, but I just wanted to point out the differences

andreaskundig:
> (After our discussion yesterday I changed my mind and think that squashing before merging to master is a good idea, so we may want to start doing this. What do you think?)

I think it's a good idea.

manifests/experiences/facebook/database.js
```diff
@@ -35,7 +37,9 @@ export default async function databaseBuilder(fileManager) {
   /// Advertisers who purchased your contact ////////////////////////////////////////////////////////////////////////////////
   db.create('advertisersContactInformation', [['name', 'TEXT']])
   const advertisersContactInformationFile = JSON.parse(
```
andreaskundig, line 39: Ok so that's an example where my suggestion does lead to uglier code, sorry

Maybe something like this looks slightly better ? Not sure it works though...

```
const advertisersContactInformationFileTexts = await fileManager.getPreprocessedTextFromId('advertisers-contact-list')
const advertisersContactInformationFile = advertisersContactInformationTexts.map(JSON.parse)[0]
```

utils/file-manager.js
```diff
@@ -288,71 +275,35 @@ export default class FileManager {
...
-    const paths = this.getFilePathsFromId(id, unique)
-    if (paths === null) {
-      return null
-    }
-    if (unique) {
-      return await this.getPreprocessedText(paths)
-    } else {
-      return await Promise.all(paths.map(p => this.getPreprocessedText(p)))
-    }
+  async getPreprocessedTextFromId(id) {
```
andreaskundig, line 294: You no longer need the second argument? That's an improvement I didn't expect.

andreaskundig (approved):
Thanks for taking my suggestions into account.

utils/file-manager.js
ffsinger: I think you may have missed my other comment https://github.com/hestiaAI/hestialabs-experiences/pull/523#discussion_r791723890

utils/file-manager.js
andreaskundig: Ok, I'm happy with my side of the trade-off, it simplifies methods from the library that will be used in many places, at the cost of giving the caller a little bit more work to make his code look good. I think he's capable of that.

JSONPath likes null better than undefined but in other places we'll prefer null. It's the caller's job to deal with things like these.

Issue #509, opened 2022-01-20
Improve the UX around Resetting or Reusing/Clearing an Experience


alexbfree:
The goal:
1. User should be able to fully reload / reset the experience (keeping same data without having to re drop it - but all views get reset to default
2. User should be able to click a button to 'wipe' the experience and send it back to "you need to upload a file" state without needing to go back to home page or reload the URL etc.
3. Current 'Cancel' button on file list/process files list is ambiguous/confusing. Either remove altogether, or change to "clear" list (prior to Process Files) and hide it after processing files is done.

Different UX solutions are possible - I am flexible. Let's keep it as simple as possible. 
Maybe two buttons - Reset Views and Clear Data

ffsinger:
Related to this, I noticed another issue:

Let's say that you load a file and run a sub-experience. Then you go back to load another file for the same company (without refreshing the page). The previously run sub-experience will not be cleared after loading the second file. You have to re-press "Run" and the visualization will be updated. This is very confusing

PR #524, opened 2022-01-25
fix: add a clear button to the experiences

fquellec:
This PR is related to #509
We are adding the ability to delete all files and return to the loading page on every experience. To do this we need to place the file manager and database in the Vuex store, many components have changed their props. This PR needs to be looked at carefully as it impacts a large part of the project. 

Demo: 
image


ffsinger:
There's a bug in the consent form. Previously we had a BaseButton that generates the zip and triggers a download. The download was done through a hidden BaseButtonDownloadData (thanks to the property `v-show="false"`). Now the button appears on the page. This code hasn't been directly changed, but it may be due to the changes in the BaseButton components.

store/index.js
```diff
@@ -51,12 +59,35 @@ export const getters = {
...
+        k,
+        Object.fromEntries(
+          manifestMap[k].defaultView?.map(b => [b.key, null]) ?? []
+        )
+      ])
+    )
+    state.fileExplorerCurrentItem = {}
+    if (state.currentDB !== null) state.currentDB.close()
+    state.currentDB = null
+    state.fileManager = null
```
ffsinger, line 89: This function could call the other mutations, I would find it cleaner:
```js
for (const company of config.experiences) {
  this.commit('setSelectedFiles', { key: company, value: [] })
  for (const experience of manifestMap[company]) {
    this.commit('setResult', { company, experience, value: null }
  }
}
this.commit('setCurrentDB', null)
this.commit('setFileManager', null)
```

This could also be turned into an action.

Also, note that currently it clears the results for **all** the experiences. I'm not sure if we want this

components/unit/consent-form/UnitConsentForm.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <VForm>
+  <VForm v-if="fileManager !== null">
```
ffsinger, line 2: General note: you can simply use `v-if="fileManager"`

store/index.js
fquellec: We already had one FileManager and one database object shared by all experiments? So there is no need to keep the states of the other experiments, right?

ffsinger: Ah yes you're right. So we should decide on the direction we wanna take, either:

1. Only keep the state of one experiment. In this case we can simplify `results` and `selectedFiles`.
1. Keep the state of the other experiments. In this case we have to store multiple `fileManager` and `db`

Let's discuss this in the meeting

ffsinger: As discussed in the meeting we don't need to keep the state of every experiment, and doing so would complicate the code so let's simplify `results` and `selectedFiles`

components/unit/files/UnitFiles.vue
```diff
@@ -183,7 +200,7 @@ export default {
     returnFiles() {
       const uppyFiles = this.uppy.getFiles().map(f => f.data)
       this.status = true
-      this.$emit('update', { uppyFiles })
+      this.$emit('update', { uppyFiles, filesEmpty: this.filesEmpty })
```
ffsinger, line 203: This value isn't used by the parent

ffsinger (approved):
I just moved the consent form to the store and now it's reset after a clear. The consent form code is still very much spaghetti but I'll fix this in a separate PR.

I think we can merge this after the conflicts are fixed

PR #529, opened 2022-01-27
Feat/json search

andreaskundig:
Replaces the old json search that cuts off any node that doesn't match, with a search that gives a list of results in a table, where a click on a row opens the tree view at the selected result.

If you need to go back to the old search, set **useOldSearch=true** in [UnitFileExplorerViewerJson.vue](https://github.com/hestiaAI/hestialabs-experiences/blob/fcf7b3c35d543208c8ae90d2a3e1cdb27780869d/components/unit/file-explorer/viewer/UnitFileExplorerViewerJson.vue#L96)

I rebased this on master, so now the tests that fail on master also fail here :/


ffsinger:
The tests that fail (`TheDataExperience.test.js`) are fixed in #528

ffsinger (commented):
It's very nice !

It would be good to automatically jump to the node when clicking on an item in the table. Otherwise we don't always notice that something happened and we have to search for the node

andreaskundig:
> It's very nice !
> 
> It would be good to automatically jump to the node when clicking on an item in the table. Otherwise we don't always notice that something happened and we have to search for the node

That's a good idea for an additional feature. Something like this would also make sense when you want to show a node as a table, now you don't necessarily notice that a table has appeared below the tree.

Issue #530, opened 2022-01-27
Recognise browser back button


alexbfree:
As an example, when you click on a file in the Summary tab, and press back, it doesn't go back to Summary, but to whatever you were viewing before the #digipower site. Can we make the back button work and go back to whatever tab / experience page you viewed last?

valentinoli:
Yes, it is possible

valentinoli:
I'll do it

valentinoli:
Should be fixed in #535

PR #535, opened 2022-01-30
Fix/auto run

valentinoli:
Addresses #365

Second attempt...


components/TheDataExperience.vue
```diff
@@ -16,19 +16,15 @@
...
-          <VTab :disabled="!success" href="#summary">Summary</VTab>
-          <VTab :disabled="!success" href="#file-explorer">Files</VTab>
           <VTab
-            v-for="(el, index) in defaultView"
+            v-for="(t, index) in tabs"
             :key="index"
-            :disabled="!success"
+            :disabled="t.disabled"
+            nuxt
+            :to="`#${t.value}`"
```
valentinoli, line 24: This part here fixes #530

components/TheDataExperience.vue
```diff
@@ -205,22 +249,30 @@ export default {
...
+      handler: debounce(function (value) {
+        this.overlay = value
+      }, 200)
     }
   },
   mounted() {
-    this.$root.$on('goToFileExplorer', () => {
-      this.tab = 'file-explorer'
-    })
+    this.switchTab('load-data')
```
valentinoli, line 269: This is to ensure that the first tab is always shown when the component is mounted...

andreaskundig (changes requested):
This looks great, thanks!

Meanwhile we have a test for TheDataExperience.vue. The changes in this PR break this test, it needs to be updated. 

I propose we fix that before merging. @ffsinger who wrote that test will have a look at it today if he has time, but it's his last day.

PR #573, opened 2022-03-11
Feat/bettertable

andreaskundig:
- a table accepts several accessors
- json viewer nodes have a button to copy accessor to clipboard
- table options syntax simplified and made more tolerant


fquellec (approved):
Nice Feature ! Just one remark, I don't think the list of accessors is working as expected. Taking the Tinder experiment as an example, how can I get a list with all the values in the same table? What is the purpose of having a list of accessors here and how can we use it?

image
 
I see two needs, one is to perform joins on different tables, we could use the accessor list here with another parameter to set the join options, and another is to create tables from deep nested structures, for this one we would need a relative JSONPATH accessor on each accessor. 

We have to define this more precisely, maybe for another PR ?

PR #596, opened 2022-03-24
Fix/various

valentinoli:


valentinoli (commented):
Card titles were being broken on characters as configured in Vuetify by default. I don't find this attractive and that's why I suggest we break on the word instead.

The tabs would not always scroll to top on tab change.

I removed redundant Vuex getters, which simply returned state properties. I changed the code to use mapState instead of mapGetters for these properties.

I saw a warning in my console that suggested to use `groupAll()` instead of `group()` in one place.

fquellec (approved):
Good to me

Issue #577, opened 2022-03-17
The SQL database does not always have time to be built before it is used. 


fquellec:
When running an experience, for example Tinder, `currentDB` is not always set before we start using it in `UnitPipelineSQL.vue`: 
<img width="447" alt="image" src="https://user-images.githubusercontent.com/17878016/158768719-25f1c323-e5f9-495c-81f5-00a22bc2bc05.png">

This causes some secondary experiences not to work properly. 

<img width="1604" alt="image" src="https://user-images.githubusercontent.com/17878016/158768971-5c194506-0330-4cef-9f06-3f794aeab204.png">

<img width="1728" alt="image" src="https://user-images.githubusercontent.com/17878016/158769094-e427a9c3-fed1-443e-b29f-0295b927c58d.png">

This bug seems to be introduced by the attempts at #365 ?

Issue #544, opened 2022-01-31
Strange issue with squished graphs when you switch tabs


alexbfree:
watch this video.
Notice what happens when I come back from Locations tab to My Activity tab the second time AFTER having Run location observations. The graphs get squished.

https://user-images.githubusercontent.com/1473244/151784435-9990a74f-6c73-4869-9680-0b82d9b7c21a.mov

valentinoli:
@fquellec we should check if this has been fixed in #535

alexbfree:
> @fquellec we should check if this has been fixed in #535

if that has already been merged/deployed then no it is still present; tested this on the newly deployed test instance today lunchtime.

alexbfree:
![Screenshot 2022-01-31 at 15 05 24](https://user-images.githubusercontent.com/1473244/151818151-42f8210b-e0c5-4bef-bc20-4a23ea22a26a.png)

Another similar weirdness seen on same page just now (on digipower instance)

fquellec:
Sorry, long time no see :) I just checked this problem, it's unfortunately introduced with the use of tabs, to define the layout, sizes of each graphic we rely on the `VRow` and `VCol` vuetify grid system, then we need a way to pass these widths to the graph library, so we define the sizes as below: 

<img width="480" alt="image" src="https://user-images.githubusercontent.com/17878016/156997911-6e8465ca-1110-4df9-b0ca-851dfb3f02e2.png">

The problem is that if the drawViz function is called before we actually render the tab, the width will be 0. This cannot be fixed by only setting the tabs to eager mode since if this only works the first time we run a file.... 
Maybe we could rerun the graphics every time we enter a tab? what do you think @valentinoli?

PR #597, opened 2022-03-25
Fix/544+577

valentinoli:
This PR addresses #544 and #577


valentinoli:
Sorry but the first 3 commits are identical to the first commits of PR #596 and I somehow accidentally added them to this branch...

Issue #526, opened 2022-01-26
Improvements to Targeting Criteria (1 Advertiser) view


alexbfree:
1) Would it be possible to show a linked table under the graph, like we do on Twitter Advertising Overview
2) Would it be possible to have an export button for CSV table data, as well as graph 

Right now this is the only place to see the targeting info for just 1 advertiser (if not in top 10) and it is hard to access the detail.

alexbfree:
we just had an example of @foucault-dumas needing this (2) for a story. Elevating

valentinoli:
@alexbfree status: test deployed

PR #627, opened 2022-04-13
fix: show table in Twitter Criteria (1 Advertiser)

valentinoli:
#526


fquellec (approved):
For another PR we should better manage the content of the tabs, the table here takes the whole width it's a bit weird, + maybe add the possibility to show/hide the data tables on each sub experience? 
image

Issue #356, opened 2021-12-01
Table UX in Generic Dated Event Viewer and Generic File Viewer


alexbfree:
Something that came up in Deep Dive. For all tables, make sure that in a long table (e.g. show all is selected), the header row is still visible even when the user scrolls down

PR #628, opened 2022-04-13
fix: improve filterable table UX

valentinoli:
* make chip selected filter deletable
* add ellipsis to text in chip of selected filter
* dynamic height depending on window height
* reduce default height when itemsPerPage === 5

#356


fquellec (approved):
Nice one !

Issue #638, opened 2022-04-19
The heatmap calendar graph should allow "only integers" for the legend.


fquellec:
When we have integer values only, the legend should not display floats: 
<img width="567" alt="image" src="https://user-images.githubusercontent.com/17878016/163962804-ea56b9a0-928e-46c8-9d7c-1f75b1640f85.png">

PR #640, opened 2022-04-19
fix: legend spacing + number of ticks using d3

fquellec:
This PR is related to #638.

Instead of manually calculating the ticks, we use the d3 scale.ticks method, if the number of ticks chosen is less than the extent, it should still display nice integer ticks. 

I also corrected the position computation of the legend.


valentinoli (approved):

valentinoli:
@fquellec you may merge if you are ok

PR #663, opened 2022-05-02
fix: Only fetch and unzip needed files in experiences

fquellec:
We previously unzip all archives in the file manager, I also added an attribute to the manifest when we need to retrieve all files, for example: for generic viewers...


valentinoli (approved):
Good :)

valentinoli:
https://github.com/hestiaAI/hestialabs/commit/415e20330c14a2ae663ab7d3019c04c2ce6dc626

Issue #708, opened 2022-05-25
Bubble Authentication


valentinoli:
Configure Nuxt to prompt the user to authenticate for access to a bubble

andreaskundig:
The password can be checked on the server with the REST API /bubbles/[bub-name]/authenticate
The password that allows to authenticate is also need to upload files with POST /bubbles/[bub-name]/files/[file-name]

valentinoli:
We need to

1. Show the list of experiences available to the current bubble
2. Redirect the user to the /login page when another bubble is accessed
3. Remove the logout function (for now)
4. Highlight the current bubble in the navigation drawer and display the current bubble

PR #721, opened 2022-06-01
Bubble authentication (client)

valentinoli:
#708


valentinoli:
One thing that is missing and I did not implement is to keep track of the user so that he is still logged in when the page is refreshed or if he leaves the session and comes back later. I think it requires a token...

fquellec (approved):
Very nice work! 
I notice a small problem when you are logged in a bubble, if you try to go to another bubble, it takes 2 clicks on the same button to get there (the first one doesn't seems to do anything). A suggestion that could also solve this problem would be to disconnect the user each time he leaves the bubble, to go to the home directory or to other pages, it is indeed a bit weird to be connected to something on the main page without knowing what. 
I will merge the PR  we can discuss this later !

valentinoli:
> the first one doesn't seems to do anything

the first click detects the user is trying to access another bubble, so the user is logged out and auto-redirected to the home page

valentinoli:
> to go to the home directory or to other pages, it is indeed a bit weird to be connected to something on the main page without knowing what

I do not think it is weird, it is a very common pattern

PR #733, opened 2022-06-07
Anonymization

wilderino:
New button in the Record tab to add some noise, in order to anonymize the data. The slider correspond to the maximum size in meters of the noise that will be added.


fquellec (commented):
Nice work !

PR #797, opened 2022-06-20
Refactor google view component

wilderino:
I created a generic map viewer component that takes a Kepler config as a parameter and will display the map with the filterable table below it. This view component can also work for aggregated experiences and you can apply k-anonymity on the aggregated data.


wilderino:
Before merging this to master I would like to change the graph for the "Public Transport" tab in the Google Agg experience. Maybe @fquellec can help me.

fquellec (approved):
Nice work ! I added the graph for public transportation.

PR #798, opened 2022-06-21
chore(eslint): fix configuration

valentinoli:


valentinoli (commented):
Fixed this error
https://stackoverflow.com/questions/68021121/fsevents-not-accessible-from-jest-haste-map

by following this advice:
https://stackoverflow.com/a/69289235/8238129

PR #844, opened 2022-07-12
fix: csv naming convention

fquellec:
A bug was introduced when forcing CSVs headers to be in pascalCase, The transformation was only made on the values and not the headers. This makes the DataValidator fails. After fixing it I had to adapt Uber, Netflix, Her and Linkedin experiences accordingly.


valentinoli (approved):

valentinoli (commented):
Remove the console.logs

PR #871, opened 2022-07-26
hestaliabs repo merged into hestialabs-experiences

andreaskundig:
To test this branch it's probably better to clone the whole repo, switch to this branch and do the npm installs. This should put in place the commit hooks.

dependabot only runs on subproject experiences. Unfortunately dependabot does not seem to be able to deal with several subrepos. 
https://github.com/dependabot/dependabot-core/issues/4993

If not, the solution will be to convert them to a true monorepo with lerna on top of all of them, or maybe try another bot than dependabot?


fquellec (approved):
Everything is working well for me, nice work ! 
We should not forget to merge the last version of hestialabs just before merging this PR!

Issue #884, opened 2022-07-27
Internationalization (for client)


andreaskundig:
A client needs the app in french. We will continue to use the same code for our app in english

Needed for a client

- [x] Define a format to store multiligual messages in packaged experiences, for example in a messages.json file that uses the same format as vue-i18n. We could keep it backwards compatible with the current format, but maybe that's not necessary.
- [x] Replace every single text in the app with a message configured in vue-i18n
- [x] Translate all the texts to french
- [x] Talk with (client) to agree on the integration with their CMS
- [x] **BUG**: the links in the menu are broken, probably because nuxt-i18n affects the behavior of the router and nuxt-link. Hopefully this can be fixed by configuring nuxt-i18n [in some way](https://i18n.nuxtjs.org/ignoring-localized-routes)

andreaskundig:
Reopening this. We found something that was not translated yet. We should at least translate everything that we need for UFC

andreaskundig:
I'm creating another issue for the rest of internationalization, beyond what that client needs

PR #895, opened 2022-07-29
feat: internationalization (i18n) (#884)

andreaskundig:
# DO NOT MERGE
This pull request is just a convenient way to document this proof of concept

This doc is copied from [I18N.md](https://github.com/hestiaAI/hestialabs-experiences/blob/feat/i18n/I18N.md)

# Goal

We want a multilingual site to which a client like UFC can add his own translations.

Translations live in these places:
- **site translations** default translations for the nuxt site: [experiences//i18n-messages-default.json](https://github.com/hestiaAI/hestialabs-experiences/blob/feat/i18n/experiences/i18n-messages-default.json)
- **experience translations**, default translations for each experience, for example in [packages/packages/google/src/index.ts](https://github.com/hestiaAI/hestialabs-experiences/blob/feat/i18n/packages/packages/google/src/index.ts#L10)
- **custom translations** loaded from an url defined in the config, for example at [experiences/static/i18n-messages-example.json](https://github.com/hestiaAI/hestialabs-experiences/blob/feat/i18n/experiences/static/i18n-messages-example.json)

**site** and **experience** translations affect different parts of the website, they should not have any common keys.
**custom translations** can override any translation of the other two.

Everything is done using [vue-i18n](https://kazupon.github.io/vue-i18n/). The override relationships between messages are implemented by modifying vue-i18n's global dictionary.

# Integration with UFC

In addition to wanting a french version of experiences, UFC wants to be able to edit the introduction text of each experience from their a cms. This solution relies on translations being loaded in a specific json format from a url.

In our first discussion with UFC, we imagined that the cms would produce html that could be included in an iframe. I hope it's possible to have it as an html snippet rather than a whole html page like iframes normally work with. It's possible that they imagined having a different url for the introduction text for each experience. If they absolutely need to do that, we could implement a function that converts their format to ours.

# Usage

Define text snippets in the vue-i18n json format. A custom translation file like  [i18n-messages-example.json](https://github.com/hestiaAI/hestialabs-experiences/blob/feat/i18n/experiences/static/i18n-messages-example.json) looks somewhat like this:
```json
{
  "en": { ... },
  "fr": {

    // This overrides a site translation
    // which stays the same for every experience
    "experience": {
      "intro": {
        "title": "Analysez vos données privées"
      }
    },

    // This overrides experience translations,
    // you'll only see it in the google experience
    "experiences": {
      "google": {
        "intro": {
          "dataPortalHtml": "Cliquez <a href=\"https://takeout.google.com/settings/takeout\">ici</a>",
          "dataPortalMessage": "<strong>Important:</strong> pour que l'expérience"
        }
      }
    }
  }
}
```

Site translations are used by calling [functions from vue-i18n like **$t()** and **$te()**](https://kazupon.github.io/vue-i18n/api/#vue-injected-methods)
``` html
  <VCardTitle class="text-h5 font-weight-bold justify-center mb-3">
    {{ $t('experience.intro.title') }}
  </VCardTitle>
```

For experience translations, keys need to be converted to the global vue-i18n namespace using a function **k()** implemented by us.
```html
  <p
    v-else-if="$te(k('dataPortalHtml'))"
    class="body-1"
    v-html="$t(k('dataPortalHtml'))"
  />
```

These examples come from [UnitIntroduction.vue](https://github.com/hestiaAI/hestialabs-experiences/blob/feat/i18n/experiences/components/unit/UnitIntroduction.vue) (The only place where they're implemented in this proof of concept)

# Implementation

The default messages are loaded at build time by [nuxt.config.js](https://github.com/hestiaAI/hestialabs-experiences/blob/88f205b92661460a96b71bb4dd3e5df631ac1c69/experiences/nuxt.config.js#L21)

The custom messages are loaded by the browser, in [experiences/layouts/default.vue](https://github.com/hestiaAI/hestialabs-experiences/blob/88f205b92661460a96b71bb4dd3e5df631ac1c69/experiences/layouts/default.vue#L99)

The experience message are loaded by components like TheDataExperience.vue and it's children. The proof of concept is implemented in [UnitIntroduction.vue](https://github.com/hestiaAI/hestialabs-experiences/blob/472ad9fcbb3b105e01c7a47d4245e3370e35eca0/experiences/components/unit/UnitIntroduction.vue#L129)

Every component needs two methods store and access its messages in the vue-i18n global dictionary
```js
    /** Convert a local, experience-specific key to a vue-i18n key. */
    // Implement this on each component with experience specific i18n
    i18nMakeGlobalKey(localKey) {
      return `experiences.${this.experienceName()}.intro.${localKey}`
    },
```
```js
    /**
     * Wrap a local, experience-specific, key-value object into a format
     * that can be merged into vue-18n's global dictionary.
     */
    i18nWrapInGlobalNamespace(messages) {
      return {
        experiences: { [this.experienceName()]: { info: messages } }
      }
    },

```

These two methods are used at component creation, when the default messages taken from the packaged experience, and then merged into the vue-i18n global dictionary:
```js
  created() {
    const experience = this.$store.getters.experience(this.$route)
    // TODO come up with a way to specify i18n messages in packaged experiences
    // const locale = this.$i18n.locale
    const defaultMessagesForLocale = pick(experience, [
      'title',
      'dataPortal',
      'dataPortalHtml',
      'dataPortalMessage'
    ])
    this.i18nMergeComponentMessages(
      defaultMessagesForLocale,
      k => this.i18nMakeGlobalKey(k),
      m => this.i18nWrapInGlobalNamespace(m),
      this.$i18n
    )
  },
```
> note: i18nMergeComponentMessages is meant to be used by all components with experience translations. It should be moved to a central place.

The config file [dev.json](https://github.com/hestiaAI/hestialabs-experiences/blob/88f205b92661460a96b71bb4dd3e5df631ac1c69/experiences/config/dev.json#L63)
has an example of a **locale** set to french and an **url** to load custom translations.

The custom translation example [configures](https://github.com/hestiaAI/hestialabs-experiences/blob/88f205b92661460a96b71bb4dd3e5df631ac1c69/experiences/static/i18n-messages-example.json#L13) the introduction of the google experience

Here's a direct link to the google experience running locally.
http://localhost:3000/experience/google#load-data

# TODO

- [x] Define a format to store multiligual messages in packaged experiences, for example in a messages.json file that uses the same format as vue-i18n. We could keep it backwards compatible with the current format, but maybe that's not necessary.
- [x] Replace every single text in the app with a message configured in vue-i18n
- [ ] Translate all the texts to french
- [ ] Talk with UFC to agree on the integration with their CMS
- [x] **BUG**: the links in the menu are broken, probably because nuxt-i18n affects the behavior of the router nuxt-link. Hopefully this can be fixed by configuring nuxt-i18n [in some way](https://i18n.nuxtjs.org/ignoring-localized-routes)


experiences/package.json
```diff
@@ -20,8 +20,8 @@
     "build:update-revision": "git rev-parse HEAD > static/git-revision.txt",
     "build:static": "nuxt generate",
     "build:server": "nuxt build",
-    "lint": "eslint --ext .js,.vue",
-    "lint:fix": "eslint --ext .js,.vue --fix",
+    "lint": "npx eslint . --ext .js,.vue",
+    "lint:fix": "npx eslint . --ext .js,.vue --fix",
```
valentinoli, line 24: Why this change?

fquellec: The previous command did not work on Andreas laptop

valentinoli:
I'm having trouble getting bubble login to work with a non-default locale

valentinoli:
> I'm having trouble getting bubble login to work with a non-default locale

Figured it out

Issue #932, opened 2022-08-15
internationalisation/globalisation issues: testing of deployment-example-experiences.netlify.app


alexbfree:
I am doing some testing on the deployment of code for a client that Andreas put at deployment-example-experiences.netlify.app. I will add places where I have found text in English or other issues by editing this ticket. Will add a comment when done.

alexbfree:
Issues found in Uber experience:

Within "Vue d'ensemble" all text within visualisations is showing in English. Specifically:

- [ ] Select data depending on the currency
- [ ] currencies in dropdown? Or are they same in all languages?
- [ ] Cumulative expenses
- [ ] Current filter: (x4)
- [ ] Total price
- [ ] Service used
- [ ] Day of week
- [ ] Begin trip address
- [ ] Values - Mon/Tue/Wed/Thur/Fri/Sat/Sun
- [ ] Values - Place names
- [ ] Values - COMPLETED and other Statuses
- [ ] All Column Headings of Table 
- [ ] General Information (in right panel)
- [ ] Orders (in right panel)
- [ ] Speed (in right panel)
- [ ] 'trips' (in right panel)
- [ ] 'mph' (in right panel)
- [ ] Total (in right panel)
- [ ] Average (in right panel)
- [ ] Miles, Min, Duration, Waiting Time, miles, min (in right panel)
- [ ] Reset all Filters (in right panel)

Under Trajets:
- [ ] In total there are 9 records of trips you have made with Uber
- [ ] (hover on the links or nodes to get more information
about the most frequent ones)
- [ ] Pickup places
- [ ] Dropoff places

Under Timeline:
(these seem to be globalised but no text being rendered for French, only language keys:

- [ ] genericDateViewer.graph-title
- [ ] genericDateViewer.from 
- [ ] genericDateViewer.to 
- [ ] genericDateViewer.found
- [ ] genericDateViewer.dated-event

The rest are not globalised it seems:

- [ ] Filename, Date, Description (column headings)
- [ ] Values - Text within description field

On the Carte de Trajets

- [ ] Entire mapbox UI is rendered in English, country names, legend etc

Tab Name:

- [ ] Location Observations (tab name is in English)
- [ ] Title: Location Observations
- [ ] See all the location events in your files, corresponding to data that has been collected on you at or concerning a specific location.

Keys that need values for French:

- [ ] genericLocationViewer.graph-title
- [ ] genericLocationViewer.found
- [ ] genericLocationViewer.location
[...]

PR #940, opened 2022-08-22
fix(i18n): #932

valentinoli:


experiences/components/chart/view/ChartViewOverviewUber.vue
```diff
@@ -38,35 +38,35 @@
...
                   <span class="filter" />
                 </span>
               </p>
             </div>
           </VCol>
           <VCol cols="12" md="4">
             <div id="week-chart">
-              <strong>Day of week</strong>
+              <strong v-t="k('Day of week')" />
               <a class="reset" style="display: none">reset</a>
```
fquellec, line 54: We have a translation key for reset too: 
`<a class="reset" style="display: none" v-t="'reset'"></a>`

fquellec (changes requested):
Nice way of handling days !

valentinoli:
Commit 7d35e92 should have the following commit message: `fix: use $n() method options to round numbers`

valentinoli (commented):
Approve

PR #949, opened 2022-08-26
feat: Add graphs for TikTok experience + dashboard generator from a table

fquellec:
We can now generate interactive dashboards from the output of the sql pipeline in the package configuration. Example of configuration: 
image


andreaskundig:
Don't have time to do a serious review before my holidays, but this look reaaally cool.

Issue #1159, opened 2022-11-14
Avoid importing from neighboring projects


andreaskundig:
Imports from other projects should always go through package.json, not just grabbing a js file from the neighboring project.

https://github.com/hestiaAI/hestialabs-experiences/blob/master/experiences/store/index.js#L2

``` javascript
import BubbleAPI from '../../data-experience/src/utils/bubble-api
```

I find 5 such imports, maybe there are more:

./experiences/store/index.js:2:import BubbleAPI from '../../data-experience/src/utils/bubble-api'
./packages/hestias-hearth/index.ts:22:import DBMS from '../../data-experience/src/utils/sql.js'
./packages/hestias-hearth/index.ts:23:import FileManagerClass from '../../data-experience/src/utils/file-manager.js'
./packages/hestias-hearth/index.ts:24:import { NodeFile } from '../../data-experience/src/utils/node-file.js'
./packages/hestias-hearth/index.ts:25:import fileManagerWorkers from '../../data-experience/src/utils/file-manager-workers.js'

valentinoli:
This one is on me

valentinoli:
I tried fixing the imports in `hestias-hearth` but it just isn't working. It is because `data-experience` doesn't have  `"type": "module"` in its `package.json`, which makes the compiler think it doesn't support ES modules. 

I will try to migrate `data-experience` to use ESM exclusively

Issue #1085, opened 2022-10-13
Remove `consent` and `publicKey` properties from deployment config docs


valentinoli:
`consent` should be moved to bubble config docs, and `publicKey` should be removed

valentinoli:
Move from this file

https://github.com/hestiaAI/hestialabs-experiences/blob/master/experiences/config/README.md

to this file

https://github.com/hestiaAI/hestialabs-bubble-server/blob/master/media-root/bubbles-config/README.md

PR #1160, opened 2022-11-14
Host data-experience on a CDN

andreaskundig:
This should not affect the rest of the code, but it's still better to test it before merging


andreaskundig:
I already published the package on the cdn.

Besides exporting TheDataExperience, It also exports some charts: chartViewBar, chartViewGenericMap and chartView, which should be able to instantiate all the others 
Here's an example of our dumbest chart, served directly from github
https://raw.githack.com/hestiaAI/hestialabs-experiences/431db2037d831d646549b9e0b40acef30c232526/data-experience/example/chart.html

See the code here
https://github.com/hestiaAI/hestialabs-experiences/blob/431db2037d831d646549b9e0b40acef30c232526/data-experience/example/chart.html

data-experience/example/index.html
```diff
@@ -1,11 +1,13 @@
 <!DOCTYPE html>
 <html>
 <head>
-  <title>data-experience</title>
+  <title>Data Experience</title>
+  <meta charset="utf-8">
```
valentinoli, line 5: Also added charset utf-8 to support utf-8 characters.

valentinoli (approved):

valentinoli:
Closes #1159 #1085 #1160

PR #1222, opened 2023-01-30
Innopark bubble

alexbfree:
adding innopark bubble (already added on bubble server)


alexbfree:
close this one as it had erroneously included some other files

PR #1248, opened 2023-02-20
Unitviewblock

andreaskundig:
This should be ready to be merged. 
I tested data-experiences and experiences with npm run dev and npm build, things seem to run.

To run the viewblock example
```
cd data-experiences
python -m http.server
```
then visit http://localhost:8000/deployment-examples/html-plain/block.html


fquellec (approved):
Shouldn't we fetch the translations `fr` and `en` from the CDN directly ? Instead of copy paste them. This is not something the end user should care about I think.

PR #1280, opened 2023-05-15
fix: issue with tabulator and boolean/objects parsing + minor changes to sar builder

fquellec:


fquellec:
Tests fails because of NPM, tested locally.

